PR #1, opened 2022-06-16
Additional Work Manager features

gcusano:


service/src/main/java/io/nuvalence/workmanager/service/domain/dynamicschema/attributes/DocumentRejectionReason.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return text;
+    }
+
+    /**
+     * Validation for Document rejection reason enum.
+     * @param text Incommoding text from json.
```
gcastro12, line 25: Incommoding should be Incoming?

service/src/main/java/io/nuvalence/workmanager/service/delegates/DetermineDistrictDelegate.java
```diff
@@ -0,0 +1,44 @@
+package io.nuvalence.workmanager.service.delegates;
```
gcastro12, line 1: This file is not really needed at all without the corresponding bpmn file/changes, but it could serve as an example of a JavaDelegate, I suppose.

jazelke (approved):
Looks good!

PR #35, opened 2023-04-08
Dsgov 360 active forms

JayStGelais:
Implements an API endpoint to get the active form configurations for a given transaction.

## How Has This Been Tested?
Existing Unit tests pass
Tested manually via postman
Due to time constraints with upcoming PTO, some automated testing has been defered. Will be picked up on [this ticket](https://dol-ewf.atlassian.net/browse/DSGOV-477)

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
```diff
@@ -54,13 +55,12 @@ public void setHttpClient(RestTemplate restTemplate) {
      * @return The user with the given id.
      */
     public Optional<User> getUserById(String userId) {
-        ResponseEntity<User> response = getJson("/api/v2/user/" + userId, User.class);
-
-        if (response.getStatusCode().is2xxSuccessful()) {
+        try {
+            ResponseEntity<User> response = getJson("/api/v2/user/" + userId, User.class);
```
asultan001, line 59: should this be /user or /users ?

swagger.yaml
```diff
@@ -569,7 +569,37 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/TransactionModel'
-  /transaction/user/all:
+  /transactions/{id}/active-forms:
```
asultan001, line 572: Is this route supposed to retrieve published forms? or what exactly do we mean by active-forms? Also, as I was thinking a bit more about this. I wonder if we should just make this `/transactions/{id}/forms` and include a query parameter to filter for only published ones?

asultan001 (approved):
@JayStGelais , this looks good to me, but just had a few none blocking questions

swagger.yaml
JayStGelais: Active here isn't related to published status, but instead what the active for that should be presented based on the state of the transaction. SUch that if we have a specific form for the current task, that is returned instead of the default form.

It's considering the active state of the transaction as input to the form selection algorithm.

service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
JayStGelais: Yes, this changed while I was out and was merged recently. I hadn't deployed the most recent changes locally so I was testing against the previous version.

Will update the PR.

swagger.yaml
asultan001: That makes sense, in that case is it expected to be a single active form or there could be multiple active forms?

service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
JayStGelais: @asultan001 

Updated and retested the user management client. 

Also saw that we had a method in there for lookup by externalId that is no longer necessary now that we include the `applicationUserId` on the token, so cleaned up the unncessary code.

asultan001 (approved):
LGTM

service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
asultan001: Great!

PR #37, opened 2023-04-12
DSGOV-368: added the external_id field to the transaction table

jgnuvalence:
DSGOV-368: added the external_id field to the transaction table, which is based on a db sequence, which is ZBase32 encoded

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/repository/TransactionRepository.java
```diff
@@ -38,4 +38,7 @@ Optional<Transaction> findByProcessInstanceId(
 
     @Query("SELECT t FROM Transaction t WHERE t.assignedTo = :userId")
     List<Transaction> searchByTransactionByAssignee(@Param("userId") String userId);
+
+    @Query(value = "SELECT nextval('transaction_sequence')", nativeQuery = true)
```
Ortega-Dan, line 42: if this query is selecting the nextval for the transaction sequence .... (question continues in next file) ....

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionFactory.java
```diff
@@ -102,6 +105,18 @@ public Transaction createTransaction(final TransactionDefinition definition)
...
                 .data(new DynamicEntity(schema))
+                .externalId(generateExternalId(sequenceValue))
                 .build();
     }
+
+    private String generateExternalId(Long sequenceNumber) {
+        if (sequenceNumber == null) {
+            return null;
+        }
+        sequenceNumber = sequenceNumber + 1; // increment the sequence value
```
Ortega-Dan, line 116: ... is this increment needed ? Just asking as nitpicking, but I believe if it is fine I am anyway approving.

Ortega-Dan (approved):
approval with context question

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
```diff
@@ -64,7 +64,8 @@ public Transaction createTransaction(final TransactionDefinition definition)
      */
     public Transaction createTransaction(final TransactionDefinition definition, String jwt)
             throws MissingSchemaException {
-        return repository.save(factory.createTransaction(definition));
+        long sequenceValue = repository.getTransactionSequenceValue();
+        return repository.save(factory.createTransaction(definition, sequenceValue));
```
asultan001, line 68: Please take my feedback with a grain of salt and I think it may be best to have @JayStGelais review this or perhaps ask one of the engineers at MPWR how that was done in their case but having the service layer be responsible for having to provide the sequence number seems like the wrong place as it requires the client to understand the internals of the domain too much and it can also lead to creating wrong transactions if the proper sequence number is not provided.  I also thought that this would have been a slightly bigger code that was encoded from a Guid but reading the architecture proposal it did seem like we wanted to do it based on a db sequence.

service/src/main/java/io/nuvalence/workmanager/service/repository/TransactionRepository.java
jgnuvalence: thanks @Ortega-Dan, I had 2 approaches and I kept one, but it seems that I did not clean up the code correctly, it is already corrected

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
jgnuvalence: Hello @asultan001, if it's okay with you we can leave this PR open until Monday, from my point of view, the service is responsible for calling the DAOs/Repositories to obtain or save the data, and also for calling the factory providing the correct information. From that point of view I think it's fine. We can have a discussion about this on Monday if that's okay with you?

asultan001: Sounds good @jgnuvalence , let's discuss more on Monday when @JayStGelais is back. I just want to avoid potential errors due to miss-use of the library. Someone reusing the same sequence number or something to that extent.

service/build.gradle
```diff
@@ -51,6 +51,7 @@ dependencies {
 	implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 	implementation 'com.vladmihalcea:hibernate-types-52:2.16.1'
 	implementation 'io.nuvalence.dsgov:dsgov-auth:0.0.11-SNAPSHOT'
+	implementation 'com.sandinh:zbase32-commons-codec_2.12:1.0.0'
```
JayStGelais, line 54: I'm nervous using this 3rd party libarary. It isn't published under any open source license and it hasn't been touched in 5 years. If we discover a bug or security vulnerability it is unlikely the maintainers will address it. 

The ZBase algorithm is vers straight forward. I think we should just implement it internally. We can use the code in this library as a reference if needed.

JayStGelais: I got this version of a function from ChatGPT. We would want to write some unit tests for it to verify it is correct.

```
public static String encodeZBase32(int number) {
    final String base32Alphabet = "ybndrfg8ejkmcpqxot1uwisza345h769";
    final int base = base32Alphabet.length();
    StringBuilder sb = new StringBuilder();
    int remainder;

    do {
        remainder = number % base;
        sb.append(base32Alphabet.charAt(remainder));
        number /= base;
    } while (number > 0);

    return sb.reverse().toString();
}
```

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionFactory.java
```diff
@@ -38,10 +40,11 @@ public class TransactionFactory {
      * Create a new transaction for a given transaction definition.
      *
      * @param definition Type of transaction to create
+     * @param sequenceValue Next value of the transaction sequence, which is the basis for creating the externalId
      * @return The newly created transaction
      * @throws MissingSchemaException if the transaction definition references a schema that does not exist
      */
-    public Transaction createTransaction(final TransactionDefinition definition)
+    public Transaction createTransaction(final TransactionDefinition definition, Long sequenceValue)
```
JayStGelais, line 47: I think we can encapsulate this logic inside the factory and not require callers to fetch the sequence number.

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
JayStGelais: Per coment above, I think moving the sequence management logic down into `TransactionFactory` makes sense.

service/src/main/resources/db/changelog/changelog-master.xml
```diff
@@ -366,4 +366,12 @@
         <dropNotNullConstraint tableName="transaction" columnName="created_by" />
         <dropNotNullConstraint tableName="transaction" columnName="subject_user_id" />
     </changeSet>
+    <changeSet id="create-transaction-sequence" author="jgiraldo">
+        <createSequence sequenceName="transaction_sequence" startValue="10" incrementBy="1" />
```
JayStGelais, line 370: Can we be a little more descriptive with this name. Maybe `transaction_external_id_sequence`. We may have future yet to be identified use cases for sequences.

JayStGelais (changes requested):

service/build.gradle
jgnuvalence: sounds good Jay, totally agreed

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
jgnuvalence: this is done

service/build.gradle
jgnuvalence: this is done

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionFactory.java
jgnuvalence: this is done

service/src/main/resources/db/changelog/changelog-master.xml
jgnuvalence: this is done

service/src/main/java/io/nuvalence/workmanager/service/utils/ZBase32Utility.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static char[] encoding = "ybndrfg8ejkmcpqxot1uwisza345h769".toCharArray();
+
+    /**
+     * Given a String is encoded inputText zbase32.
+     *
+     * @param inputText String to be base32 encrypted
+     * @return encrypted base64 String
+     */
+    public static String encode(String inputText) {
```
JayStGelais, line 18: This encoder should be taking integers (should probably support both int and long.

This logic is also very difficult to follow. 

Let me take a stab at implementing this. I think we can make this much simpler. I would use bitmasks and bit shifting to pull out the 5 bit chunks of integer data.

JayStGelais (changes requested):

JayStGelais (approved):

JayStGelais (approved):

PR #71, opened 2023-04-27
DSG-14: add dsgov-logging lib to handle the loging app config

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


service/build.gradle
```diff
@@ -83,7 +84,7 @@ tasks.withType(JavaCompile) {
 
 jib {
 	container {
-		jvmFlags = ['-Dspring.profiles.active=deployment']
+		jvmFlags = ['-Dspring.profiles.active=local']
```
dtsong, line 87: I think we want to be a little careful about defaulting this to `local` because originally, the `application-deployment.yaml` will capture the configuration values that will be needed in the GKE deployment.

I think the way that this works is that when we do a `./gradlew build jib`, it will utilize the local profile when building that image.

jgnuvalence: I paired with @dtsong and made changes accordingly

dtsong: Things look good, ready to test in the cloud deployment!

dtsong (approved):

PR #91, opened 2023-05-12
Feature/dsg 629/add my transactions endpoint

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -125,31 +128,24 @@ public ResponseEntity<Map<String, FormConfigurationRenderModel>> getTransactionA
...
-            throw new ForbiddenException();
-        } catch (ForbiddenException e) {
-            log.error(AUTH_NOT_SET_ERR_MESSAGE);
-            return ResponseEntity.status(HttpStatus.FORBIDDEN).build();
+        final Page<TransactionModel> transactions =
+                service.getTransactionsByUser(userId, sortBy, sortOrder, page, size)
+                        .map(mapper::transactionToTransactionModel);
+
+        if (transactions.isEmpty()) {
+            throw new NotFoundException();
```
asultan001, line 145: I don't think we want to return an exception in this case as it is likely that a user doesn't have any transactions when they are first starting to use the system

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -446,4 +442,24 @@ private List<String> getAssignedToList(List<String> assignedTo, Boolean assigned
         }
         return assignedTo;
     }
+
+    private String getAuthenticatedUserId() {
```
asultan001, line 446: It seems like this would be a good utility in our Auth utils library. I can see every service wanting a utility method like this and it would allow us to test this well.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -446,4 +442,24 @@ private List<String> getAssignedToList(List<String> assignedTo, Boolean assigned
...
+            if (authentication instanceof UserToken) {
+                final UserToken token = (UserToken) authentication;
+                return token.getApplicationUserId();
+            }
+        }
+        log.error(AUTH_NOT_SET_ERR_MESSAGE);
+        throw new ForbiddenException();
+    }
+
+    private PagedTransactionModel generatePagedTransactionModel(
```
asultan001, line 458: I wonder if it makes sense to try to extract this into a utility library as well since every service that requires returning paged results will have to do something similar.

swagger.yaml
```diff
@@ -627,20 +627,50 @@ paths:
                   $ref: "#/definitions/FormConfigurationRenderModel"
         '404':
           description: A transaction with the given ID does not exist.
-  /transactions/user/all:
+  /my-transactions:
     get:
-      summary: Get all transactions for a given user.
-      operationId: getTransactionsByUser
+      summary: Get all transactions for authenticated user.
```
asultan001, line 632: I would reword this as: 'Get all transactions created by the authenticated user.' just to clarify that these are not transactions assigned to the current agent or something like that.

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
```diff
@@ -327,11 +327,19 @@ public List<TransactionCountByStatusModel> getTransactionCountsByStatus(
...
+     * @param sortOrder Determines direction for resulting sorting
+     * @param page      The page number to show transactions for
+     * @param size      The number of transactions to retrieve
      * @return List of transactions
      */
-    public List<Transaction> getTransactionsByUser(String userId) {
-        return repository.searchByTransactionByUser(userId);
+    public Page<Transaction> getTransactionsByUser(
+            String userId, String sortBy, String sortOrder, Integer page, Integer size) {
+        return repository.findByCreatedByOrSubjectUserId(
```
asultan001, line 339: Why do we need a specific repository method for this? couldn't we use the previous method that allows for searching of transactions? I would think that this is a subset of what we already supported before. Ideally I want to avoid having multiple search methods if possible.

jgnuvalence: I was thinking about this, but our findAll method that is used for the Transactions search does an AND between the search params, in this case between CreatedBy and SubjectUserId, but in our case we need an OR

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jgnuvalence: Since this is exclusive to Transactions, it seems to me that this is a good place to have this method, don't you think?

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jgnuvalence: I created a new PR to move this to dsgov-auth:
[PR](https://github.com/Nuvalence/dsgov-libs/pull/12)

swagger.yaml
jgnuvalence: In the query I saw that this is for transactions created by the user or in which this is the subject:

`t.createdBy = :userId OR t.subjectUserId = :userId`

 I don't really know what this means, but I want to confirm that by mentioning this you still think that this should be the summary?

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
asultan001: Yeah, right now this is transaction specific but I was thinking that we could create a Generics method that would be consistent and applicable to every service needing to return a paged result for example something like this:

`
   PageModel<Y> generatePagedModel<T>(Page<T> records) {
        PageModel<Y> model = new PageModel<Y>(); 
        model.items(records.toList());
        model.setPagingMetadata(pagingMetadataMapper.toPagingMetadata(records));
        return model;
    }
`
Or something to that extent. It is not necessary at this time but just an idea.

swagger.yaml
asultan001: Actually that is a good question, I don't know the domain well so I'm not sure what that means. @JayStGelais can you clarify if that is correct? What does subjectUserId mean in this context?

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
asultan001: hmm got it, yeah that's a bit unfortunate. I wonder if we should enhance this method now or if it will make it to complex to consume. @JayStGelais what do you think? My concern is that some of the additional filtering criteria can still be useful over time so we will constantly be having to update the search capability.

swagger.yaml
JayStGelais: Created By and Subject are 2 different tracked user references because we will want to support transactions created by a 3rd party agent, but pertaining to a different user.

Use cases could include transaction entered into the system by:
- agency user, transcribing a paper application
- Allowed proxy users, such as legal counsel
- Transactions entered by a parent or guardian of a minor and later transferred to the subject as an adult.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jgnuvalence: I think this is interesting, the problem I'm seeing now is that `PagedTransactionModel` and `TransactionModel` are classes generated by openAPI, I've been researching how to make a `PageModel<Y>` generated by openAPI and I find it very complicated, if you think this can be sufficiently valuable I could create a jira ticket for that, makes sense?

asultan001: Got it, in that case maybe we leave it as is. No point in trading complexity for a simple utility method. 
Abe

swagger.yaml
asultan001: Got it, in that case a more appropriate description would be : 'Get all transactions created by or on behalf of the authenticated user.'

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
JayStGelais: I think we can get a more flexible search feature if we use search by `Specification`.

https://reflectoring.io/spring-data-specifications/

jgnuvalence: I understand the reason for using specificatio pattern in `findAll` wich is out Transaction reach query method, but I don't see any benefit in doing it for this method, which is so specific, don't you think. @JayStGelais ? As I mentioned in the case of `findAll`, currently each parameter generates an AND and not an OR, which is what we need in this specific case

swagger.yaml
```diff
@@ -627,20 +627,50 @@ paths:
...
+      operationId: getTransactionsForAuthenticatedUser
       tags: [ 'Transaction' ]
+      parameters:
```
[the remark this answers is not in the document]
jgnuvalence: Thanks Abe, it looks much better like this, I just pushed this change

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jgnuvalence: ok I just removed that

swagger.yaml
jgnuvalence: sure, I just added that summary

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
JayStGelais: The choice to join query clauses with AND or OR is a decision made in code constructing the query. I think the `findAll` repository method is sufficient for this use case, but we should have a dedicated service method for this that constucts the query appropriately.

jgnuvalence: @JayStGelais @asultan001 I just implemented the solution discussed during the standup

service/src/main/java/io/nuvalence/workmanager/service/models/BaseFilters.java
```diff
@@ -12,4 +21,35 @@ abstract class BaseFilters {
     private String sortOrder;
     private Integer page;
     private Integer size;
+
+    public abstract Specification<Transaction> getTransactionSpecifications();
```
asultan001, line 25: I wonder if the TransactionFilters class is a better home for this given it is Transaction specific.

jgnuvalence: I totally agree @asultan001 , I just pushed the change

asultan001 (approved):
LGTM

PR #93, opened 2023-05-16
Bug/dsg 696/add arraylist to localdate converter

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


jgnuvalence:
It is expected that for now the checks of this PR will fail since version `0.1.2` of `dsgov-auth` does not exist yet, this PR must first be merged:
[](https://github.com/Nuvalence/dsgov-libs/pull/14)

and must also release the new version

service/src/main/java/io/nuvalence/workmanager/service/domain/dynamicschema/DataConversionSupport.java
```diff
@@ -193,4 +199,16 @@ private static Document convertMapToDocument(Map<String, Object> documentMap) {
                 .questionKey(questionKey)
                 .build();
     }
+
+    private static LocalDate convertArrayListToLocalDate(ArrayList<Integer> value) {
+        final int dateArraySize = 3;
+        if (value.size() < dateArraySize) {
```
asultan001, line 205: should this be !=  instead? do we ever expect more values to be passed in the array? Maybe time related values would be sent?

service/src/test/java/io/nuvalence/workmanager/service/domain/dynamicschema/DataConversionSupportTest.java
```diff
@@ -35,6 +37,22 @@ void convertStringToLocalTime() {
         assertNull(DataConversionSupport.convert("", LocalTime.class));
     }
 
+    @Test
+    void convertArrayListToLocalDate() {
```
asultan001, line 41: Can we add some negative tests here also? what happens when negative numbers are sent or days and months are greater than 31 and 12? what happens when you try to add a day that doesn't exist on a month like Feb 31st? etc.

jgnuvalence: @asultan001  this arrayList is created by jackson when saving a LocalDate as json, this is not an array created by us, so this is not supposed to happen

service/src/main/java/io/nuvalence/workmanager/service/domain/dynamicschema/DataConversionSupport.java
jgnuvalence: I just pushed a fix for this @asultan001

service/src/test/java/io/nuvalence/workmanager/service/domain/dynamicschema/DataConversionSupportTest.java
asultan001: Got it. In that case then it probably doesn't require that level of validation. My thought though was that given this is an API, anyone could call the API with any info correct? or is this being converted after it reaches the API?

service/src/main/java/io/nuvalence/workmanager/service/domain/dynamicschema/DataConversionSupport.java
```diff
@@ -193,4 +200,22 @@ private static Document convertMapToDocument(Map<String, Object> documentMap) {
...
     }
+
+    private static LocalDate convertArrayListToLocalDate(ArrayList<Integer> value) {
+        try {
+            final int dateArraySize = 3;
+            if (value.size() != dateArraySize || value.contains(null)) {
+                throw new DateTimeException("Malformed date");
+            }
+
+            int year = value.get(0);
```
JayStGelais, line 211: Why are we useing a list of Integers to submit data data.

Can't the front-end send a string in YYYY-MM-DD format?

PR #135, opened 2023-06-14
Feature/dsg 781/record transaction note

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/NoteApiDelegateImpl.java
```diff
@@ -54,13 +54,23 @@ public ResponseEntity<List<NoteModelResponse>> getTransactionNotes(UUID id) {
...
             throw new ForbiddenException();
         }
+
         NoteModelResponse noteResponse =
                 noteMapper.noteToNoteModelResponse(
                         noteService.createTransactionNote(
                                 id, noteMapper.noteModelRequestToNote(noteModelRequest)));
+
+        try {
+            noteService.postAuditEventForTransactionNote(
```
JayStGelais, line 68: Instead of mapping the response back to a note, you could capture the return value of `noteService.createTransactionNote` to a local variable and pass that here. The double mapping here increases the potential for future bugs.

service/src/main/java/io/nuvalence/workmanager/service/models/NoteAddedAuditEventDto.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Data for an audit event regarding transaction note creation.
+ */
+@Data
+@AllArgsConstructor
+@Slf4j
+public class NoteAddedAuditEventDto {
```
JayStGelais, line 15: In regards to packaging, could we put audit event data in a separate package to make it clear that it is different than the main business models? Maybe an `auditevents` sub-package under models?

service/src/main/java/io/nuvalence/workmanager/service/service/NoteService.java
```diff
@@ -85,6 +101,59 @@ public Note createTransactionNote(UUID transactionId, final Note note) {
...
+     * Posts an audit event for transaction notes being added.
+     *
+     * @param transactionId transaction identifier.
+     * @param note added note.
+     * @return id of posted audit event.
+     * @throws ApiException result from request to audit service.
+     */
+    public AuditEventId postAuditEventForTransactionNote(UUID transactionId, Note note)
+            throws ApiException {
+        Optional<Transaction> optionalTransaction =
```
JayStGelais, line 118: You can save yourself a read of the transaction by using the created timestamp on the note for the event timestamp. I'm not sure that we are even updating lastupdatedTimestamp when we add a note, as it isn't an pdate to the actual transaction.

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/controllers/NoteApiDelegateImpl.java
jcsuarezm: Thank you Jay. The double mapping has been eliminated.

service/src/main/java/io/nuvalence/workmanager/service/models/NoteAddedAuditEventDto.java
jcsuarezm: Thank you Jay. The class was moved to the auditevents sub-package.

service/src/main/java/io/nuvalence/workmanager/service/service/NoteService.java
jcsuarezm: Thank you Jay. The timestamp is now obtained from the note.

service/src/main/java/io/nuvalence/workmanager/service/controllers/NoteApiDelegateImpl.java
```diff
@@ -54,13 +55,23 @@ public ResponseEntity<List<NoteModelResponse>> getTransactionNotes(UUID id) {
...
+        Note note =
+                noteService.createTransactionNote(
+                        id, noteMapper.noteModelRequestToNote(noteModelRequest));
+
+        NoteModelResponse noteResponse = noteMapper.noteToNoteModelResponse(note);
+
+        try {
+            noteService.postAuditEventForTransactionNote(id, note);
+        } catch (Exception e) {
+            log.error("Audit event was not published with exception {}", e);
```
JayStGelais, line 72: Can we add more context to this message so that we have necessary info in the logs for remedy. I think we would need the transactionId, noteId and the fact this would have been a note created event.

Also, it's better to use the version of `log.error` that takes an exception as the 2nd argument instead of using string interpolation, because then it will properly marshall the stacktrace to JSON.

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/controllers/NoteApiDelegateImpl.java
jcsuarezm: Hi Jay, thank you. I have addressed the logging comment and added a test to validate the message is correct.

JayStGelais (approved):

PR #141, opened 2023-06-20
Camunda managed Transaction submittedOn property

Ortega-Dan:
[DSG-904: Camunda managed submittedOn transaction property](https://dol-ewf.atlassian.net/browse/DSG-904)

## Description

Automatic setting of submittedOn transaction property by the server based on workflow config.

## Motivation and Context

Track the time a transaction is submitted.

## How Has This Been Tested?

Local testing with API and front-end and unit tests.

## Types of changes

Patch

Minor

Major

## Checklist:

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly


service/src/main/java/io/nuvalence/workmanager/service/service/TransactionFactory.java
```diff
@@ -76,7 +76,7 @@ public Transaction createTransaction(final TransactionDefinition definition)
                 .transactionDefinitionKey(definition.getKey())
                 .priority("medium") // default to medium TODO: check with FE dropdown to ingest that
                 .createdBy(createdByUserId)
-                // processInstanceId y status will be set in TransactionStatusUpdateDelegate, an
+                // processInstanceId and status will be set in TransactionStatusUpdateDelegate, an
```
jgnuvalence, line 79: thank you for this Dan :)

jgnuvalence (approved):

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionSubmitDelegate.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public void execute(DelegateExecution execution) throws Exception {
+        UUID transactionId = (UUID) execution.getVariable(TRANSACTION_ID_KEY);
+
+        if (transactionId != null) {
+            Optional<Transaction> transactionOptional =
+                    transactionService.getTransactionById(transactionId);
+            if (transactionOptional.isPresent()) {
+                Transaction transaction = transactionOptional.get();
+                transaction.setSubmittedOn(OffsetDateTime.now());
```
jgnuvalence, line 32: How are you making sure that the submittedOn has the same value as the lastUpdatedTimestamp here?

jgnuvalence (changes requested):
I think you are missing a logic to make sure that submittedOn and lastUpdatedTimestamp have the exact same value, please correct me if I'm wrong

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionSubmitDelegate.java
Ortega-Dan: Thank you for the check Julian, I believe it is not really a hard requirement since both are set at different times, but the lastUpdatedTimestamp gets updated on the save action triggered here in line 33, but captured by the UpdateTrackedEntityEventListener [here](https://github.com/Nuvalence/dsgov-work-manager/blob/3765545efebb8e0e6b48bb7c3a749f5be7ba6547/service/src/main/java/io/nuvalence/workmanager/service/domain/UpdateTrackedEntityEventListener.java#L39-L45). 
So those are in general just some nano, milis, or at most a couple of seconds away.
I anyway ping @JayStGelais for confirmation about the requirement.

JayStGelais: I think we need to ensure they are kept in sync. Can we make sure that we apply the update to `lastUpdatedTimestamp`, prior to completing workflow tasks. Then we can have this line copy the `lastUpdatedTimestamp` to `submittedOn`

Ortega-Dan: mm, nice. Thank you guys. I really didn't see it as a hard requirement but I can see there can be a use case on this, as a way to prove the last thing done at that time was the actual submit, and possibly other use cases. I'll check on it then.

JayStGelais: I think we should take management of `lastUpdatedTimestamp` out of the JPA Event Listener, so that we can have access to the updated value prior to the flush of the Persistence Manager. This will also be important for consistent timestamps with audit events.

service/src/main/java/io/nuvalence/workmanager/service/domain/UpdateTrackedEntityEventListener.java
```diff
@@ -32,15 +32,12 @@ public void preUpdateTrackedEntityPersist(final UpdateTrackedEntity entity) {
...
     /**
-     * Sets last updated tracking data for the given entity.
+     * Sets last_updated_by user for the given entity.
      *
      * @param entity entity to update
      */
     @PreUpdate
     public void preUpdateTrackedEntityUpdate(final UpdateTrackedEntity entity) {
         UserUtility.getCurrentApplicationUserId().ifPresent(entity::setLastUpdatedBy);
-
```
JayStGelais, line 42: I don't think we should have moved the update to the date out of the listener, We should have just taken this listener off of transactions and manged these fields directly in the update method.

Sorry my last comment wasn't more clear.

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
```diff
@@ -118,6 +119,25 @@ public Optional<Transaction> getTransactionByProcessInstanceId(String processIns
...
+     * <p>The persisted transaction object returned will have the submittedOn and the lastUpdatedTimestamp 
+     * attributes set to the same value after calling this method.
+     *
+     * @param transaction Transaction containing updated data.
+     * @return Transaction post-update
+     */
+    public Transaction submitTransaction(final Transaction transaction) {
+        final OffsetDateTime now = OffsetDateTime.now();
+        transaction.setSubmittedOn(now);
+        transaction.setLastUpdatedTimestamp(now);
```
JayStGelais, line 140: This will lead to the lastUpdatedTimestamp being different than the timestamp on the audit events from this action. 

In the delegate we should just do this:

`transaction.setSubmittedOn(transaction.getLastUpdatedTimestamp());`

service/src/main/java/io/nuvalence/workmanager/service/domain/UpdateTrackedEntityEventListener.java
```diff
@@ -39,8 +43,19 @@ public void preUpdateTrackedEntityPersist(final UpdateTrackedEntity entity) {
...
 
-        final OffsetDateTime now = OffsetDateTime.now();
-        entity.setLastUpdatedTimestamp(now);
+    private OffsetDateTime getCommonTimestamp() {
+        OffsetDateTime currentTimestamp;
+        try {
+            currentTimestamp =
+                    SpringApplicationContext.getBeanByClass(RequestContextTimestamp.class)
+                            .getCurrentTimestamp();
+        } catch (BeansException e) {
```
JayStGelais, line 56: We should probably log a warning here if this actually happens.

JayStGelais (approved):

jgnuvalence (approved):

PR #142, opened 2023-06-21
Feature/dsg 780/record transaction status changed audit event

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


asultan001:
@jgnuvalence for some reason, SonarQube is identifying your branch with the following bug but I don't think you changed that code. Regardless, we should probably fix it:

image

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/StatusChangedAuditHandler.java
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            String eventSummary;
+            if (before != null && !before.equals(after)) {
+                eventSummary =
+                        String.format(
+                                "At %s, transaction %s changed its status to [%s]. Previously it"
+                                        + " was [%s]",
+                                eventTime, transactionExternalId, after, before);
+            } else {
+                return;
```
Ortega-Dan, line 54: I know we assume no status should be null, but I believe we should also generate an event when the status before was null but after is not null. As a way to confirm an actual change of status.

jgnuvalence: do you mean in the case there is an error or something like this, because this scenario should not happen, every transaction should be created with a status

service/src/test/java/io/nuvalence/workmanager/service/audit/StatusChangedAuditHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+        auditHandler.handlePostUpdateState(transaction);
+
+        OffsetDateTime eventTime = OffsetDateTime.now();
+        String originatorId = "originatorId";
+        auditHandler.publishAuditEvent(eventTime, originatorId);
+
+        verify(transactionAuditEventService)
+                .postStateChangeEvent(
+                        eq(originatorId),
+                        eq(originatorId),
```
jcsuarezm, line 51: Hi, do you believe that this use case would benefit from originatorId and userId being different? I believe there could be situations in which the one who updates is different from the user.

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/StatusChangedAuditHandler.java
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            transactionAuditEventService.postStateChangeEvent(
+                    originatorId,
+                    originatorId,
```
Ortega-Dan, line 59: Also wonder if we should also track the user as the user owner of the transaction, since an agency user may change the state of the transaction but the user would be the one that submitted it.

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/StatusChangedAuditHandler.java
jgnuvalence: ok cool, thank you for checking

Ortega-Dan (commented):
I mainly think about if we should also track the user, as Juan also pointed distinct to the originator of that event, like when a transaction from a public user is approved or denied by an agency user
image

service/src/test/java/io/nuvalence/workmanager/service/audit/StatusChangedAuditHandlerTest.java
jgnuvalence: totally agreed, but I have a doubt, in this case the userId should be the createdBy or the subject_userId ?, I would like @asultan001 to answer this, since this is supposed to be the affected user

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/StatusChangedAuditHandler.java
jgnuvalence: sure, left a comment in the above thread asking Abe about this. thank you

service/src/test/java/io/nuvalence/workmanager/service/audit/StatusChangedAuditHandlerTest.java
jgnuvalence: just committed a fix for this

jcsuarezm (approved):

PR #144, opened 2023-06-21
feature/DSG-783/audit-transaction-submitted

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Delegate to report audit events regarding transactions being submitted.
+ */
+@RequiredArgsConstructor
+@Component("SimpleTransactionAuditEventDelegate")
+@Slf4j
+public class SimpleTransactionAuditEventDelegate implements JavaDelegate {
```
Ortega-Dan, line 25: Not sure but wonder if the name of this class may be more specific to what it does. Also if this is to handle submit events, this other class also exists [there](https://github.com/Nuvalence/dsgov-work-manager/blob/109e0c7aacf010b69ebad0de38391cea2509401d/service/src/main/java/io/nuvalence/workmanager/service/service/TransactionSubmitDelegate.java#L1-L42). Could we better use that one?. Triggering the discussion. What do you thing ?

Ortega-Dan (commented):
Two questions in the single comment about the new Delegate

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
jcsuarezm: Hi @Ortega-Dan, thank you. I believe the name was a design decision, as it was specified in the ticket's description, but I would also prefer using a different name.  @asultan001 @JayStGelais, please let me know if you would like a different name to be used or if it would be better to keep the current one.  
image
Yes there are some changes in demo-config, found in this PR: https://github.com/Nuvalence/dsgov-demo-configurations/pull/14/files

Ortega-Dan: Besides the name of the class, I also would like to raise the question about using the [existing Delegate](https://github.com/Nuvalence/dsgov-work-manager/blob/109e0c7aacf010b69ebad0de38391cea2509401d/service/src/main/java/io/nuvalence/workmanager/service/service/TransactionSubmitDelegate.java#L1-L42) mentioned that handles submit events already. Setting the submittedOn timestamp and then sending a notification for a Delegate that handles actions for that event seem to be cohesive.

asultan001: hmm, so I'm not familiar with the design decision of why a new delegate was needed but I suspect that if the existing delegate is able to do all the work, we should just reuse the existing delegate. @JayStGelais probably needs to provide  some insights here but again if the current delegate is able to support everything we need, I don't see a need for a new one.

asultan001: As I'm reading this again, I'm not sure if submitting a transaction and sending a notification is desired for every audit event so it probably does make sense to separate that into different delegates but I'll let @JayStGelais answer that to be certain.

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
```diff
@@ -0,0 +1,69 @@
...
+                    transactionService.getTransactionById(transactionId);
+            Optional<String> activityEventType =
+                    CamundaPropertiesUtils.getExtensionProperty("activity_event_type", execution);
+            if (optionalTransaction.isEmpty() || activityEventType.isEmpty()) {
+                return;
+            }
+
+            AuditActivityType auditActivityType =
+                    AuditActivityType.fromValue(activityEventType.get());
+            String summary =
```
JayStGelais, line 49: Might make sense to make the summary configurable, perhaps using a handlebars syntax.

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/AssignedToChangedAuditHandler.java
```diff
@@ -66,7 +65,7 @@ public void publishAuditEvent(OffsetDateTime eventTime, String originatorId) {
 
             transactionAuditEventService.postStateChangeEvent(
                     originatorId,
-                    originatorId,
+                    subjectUserId,
```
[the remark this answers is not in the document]
jcsuarezm: Thanks @JayStGelais, these were set to be both the id from the token, for this particular case.

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
jcsuarezm: Handlerbars was introduced for this purpose.

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
jcsuarezm: They have been kept separate, thank you!

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
```diff
@@ -0,0 +1,77 @@
...
+            if (optionalTransaction.isEmpty() || activityEventType.isEmpty()) {
+                return;
+            }
+
+            AuditActivityType auditActivityType =
+                    AuditActivityType.fromValue(activityEventType.get());
+
+            Handlebars handlebars = new Handlebars();
+            String templateString =
+                    "Recording audit event for transaction with id {{transactionId}}.";
```
JayStGelais, line 56: What I meant here regarding a handle bars capable template, was to pull this string from an extension property, which may include handlebar refernces to a set of available inputs.

makes sense to include transactionId, externalId, status, etc...

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/service/SimpleTransactionAuditEventDelegate.java
jcsuarezm: Thank you @JayStGelais, will do that now.

jcsuarezm: The change has been implemented @JayStGelais

JayStGelais (approved):

PR #150, opened 2023-06-30
Feature/dsg 1026/enhance notes api

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/models/BaseFilters.java
```diff
@@ -25,25 +24,16 @@ abstract class BaseFilters {
...
-        Sort sort = Sort.unsorted();
+        Sort sort;
+
+        if (sortBy.equals("noteType")) {
+            sortBy = "noteType.name";
+        }
 
-        // TODO: These are the only valid sorts so far, replace later when columns are
-        // more defined
-        switch (sortBy.toLowerCase(Locale.ENGLISH)) {
```
JayStGelais, line 32: Wont this break transaction sorting? We probably need separate filter implementations for the different searchable entities.

service/src/main/java/io/nuvalence/workmanager/service/models/TransactionNoteFilters.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+/**
+ * Filters to get transaction notes by.
+ */
+@Getter
+@Setter
+public abstract class TransactionNoteFilters extends BaseFilters {
```
JayStGelais, line 16: Why is there a subclass of this? Why not just include the implementation from `SearchTransactionNoteFilters` in this class? 

Not saying this is wrong, I just don't understand what we're planning for since, `SearchTransactionNoteFilters` is the only concrete implementation of this so far.

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/models/BaseFilters.java
jcsuarezm: Hi @JayStGelais, the new code should be sufficient to make sure that transaction sorting is working as expected. I tested this, as shown here: 
image
image
image
image
I believe the implementation is flexible enough to maintain a single class, but please let me know if you believe it would be better to make this different for each searchable entity.

service/src/main/java/io/nuvalence/workmanager/service/models/TransactionNoteFilters.java
jcsuarezm: Thank you Jay, the classes were unified to simplify the filtering implementation.

JayStGelais (approved):

PR #172, opened 2023-07-25
[dsgov-work-manager]Engagement Accelerator: Parameterize application source code and configs [DSG-1210]

adminco:
This PR remotes static configuration values in order to accelerate the process of bootstrapping the application, (its services) and its environment for a new client engagement; Nuvalence-specific infrastructure references would be pulled out to specific contexts that are clear configuration spots such as:

Helm chart values.
Skaffold profiles.
Cloud Build custom substitutions (provided by Terraform).

Reference:
https://dol-ewf.atlassian.net/browse/DSG-1210
https://github.com/Nuvalence/dsgov-audit-service/pull/58


acceptance-tests/src/functionalTest/java/io/nuvalence/platform/utils/cucumber/contexts/ScenarioContext.java
```diff
@@ -19,9 +18,7 @@
 @Setter
 @Slf4j
 public class ScenarioContext {
-    private static final String baseUri =
```
JayStGelais, line 22: Can we keep this default for local devlopment. It makes the tests easier to run as part of the developer workflow.

audit-service-client/src/main/java/io/nuvalence/workmanager/auditservice/client/ApiClient.java
```diff
@@ -65,6 +65,8 @@ public class ApiClient {
   private Duration readTimeout;
   private Duration connectTimeout;
 
+  private static final String defaultBaseUri = System.getenv("DEFAULT_SERVICE_URI");
```
JayStGelais, line 68: This shouldn't be using the same variable as the work manager service API in the acceptance tests as this will need to refer to the audit service API.

JayStGelais (changes requested):

acceptance-tests/src/functionalTest/java/io/nuvalence/platform/utils/cucumber/contexts/ScenarioContext.java
dtsong: Given that `api.dsgov.test` gets created in a developer's laptop via the `dsgov-local-environment` script where it maps the /etc/hosts to it, think it is ok to leave this in, happy to provide context if needed @adminco.

audit-service-client/src/main/java/io/nuvalence/workmanager/auditservice/client/ApiClient.java
dtsong: Think it would be best that we set this to the `api.dsgov.test/as` URL with the expectation to provision Audit Service in the local Minikube cluster if wanting to execute this test. And in a situation where a developer wishes to test against the cloud version, that can be configured via a specific environment variable.

dtsong: Think we should call this `defaultBaseUri = System.getenv("AUDIT_SERVICE_BASE_URL")` for clarity on which variable ought to be set.

dtsong: Upon trying this locally, it looks like maybe we will want to do something like this for the time being:

`defaultBaseUri = Optional.ofNullable(System.getenv("SERVICE_URI")).orElse("https://api-dev.dsgov.demo.nuvalence.io/as");`

That said, I am concerned about a real client engagement scenario where this value is not provisioned yet and it basically couples Work Manager and Audit Service to be running at the same time in order to have a successful `./gradlew clean build`.

acceptance-tests/src/functionalTest/java/io/nuvalence/platform/utils/cucumber/contexts/ScenarioContext.java
adminco: Okay no worries, I have reverted the change.

dtsong (approved):

audit-service-client/src/main/java/io/nuvalence/workmanager/auditservice/client/ApiClient.java
```diff
@@ -65,6 +66,8 @@ public class ApiClient {
   private Duration readTimeout;
   private Duration connectTimeout;
 
+  private static String defaultBaseUri = Optional.ofNullable(System.getenv("SERVICE_URI")).orElse("https://api-dev.dsgov.demo.nuvalence.io/as");
```
JayStGelais, line 69: I think we can leave this file as us before this change set. This file can and will be updated programmatically by a gradle task and any we should expect changes we make here to be overwritten by that process.

adminco: @JayStGelais Updated.

JayStGelais (approved):

PR #202, opened 2023-08-22
Bug/dsg 1326/formio validation changes

jcsuarezm:
## Description
Backend:

- Refactored the update transaction endpoint to find form configuration key via task id, instead of being received directly from the frontend. 

- Added completeTask as a query param, as now the presence of taskId is not only to be used to complete Camunda tasks.

- Added context to the request body, as it might be used to find the form configuration key.

- Modified formio validation error creation to match the front end’s structure. 

- Modified tests to respond to the new api endpoint. 

Front end:

- Refactored the call to the update transaction endpoint to include context, completeTask and formStepKey, which triggers step validation. 

- Mapped possible formio validation errors to the structure used in the front end. 

- Employed the same method as front end error handling to show these validation errors.

- Modified tests to respond to the new api endpoint. 

To make sure that no functionality was affected by these changes, a full transaction submission process was tested sucessfully, showing how the new api works for situations in which the Camunda is not to be completed, as well as situations in which the task is to be completed. (Screenshots #  1 and # 2). To test that validation was indeed triggered, a faulty endpoint was used, in which validation errors were triggered on purpose (Screenshot # 3)

1. 
image
2. 
image
3. 
image

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
```diff
@@ -847,50 +850,62 @@ public Map<String, Object> unifyAttributeMaps(
...
+
+        final TransactionDefinition transactionDefinition =
+                transactionDefinitionService
+                        .getTransactionDefinitionById(transaction.getTransactionDefinitionId())
+                        .orElseThrow();
+
+        String formKey = transactionDefinition.getDefaultFormConfigurationKey();
+        if (task != null) {
+            Optional<String> optionalFormConfigurationKey =
+                    transactionDefinition.getFormConfigurationKey(task, userType, context);
```
Ortega-Dan, line 876: the getFormConfigurationKey already returns the default one if nothing matches. And it has unit tests checking that. So this could do it this way or the other. (So this is just nitpicking, not needing change if not wanted).

service/src/main/java/io/nuvalence/workmanager/service/service/FormConfigurationService.java
```diff
@@ -198,10 +198,18 @@ private void addConfigErrorMessage(
...
+            if (controlNameParts.length > 0) {
+                label = controlNameParts[controlNameParts.length - 1];
+            }
+        }
         NuvalenceFormioValidationExItem validationError =
                 NuvalenceFormioValidationExItem.builder()
                         .controlName(componentKey)
-                        .errorMessage(message)
+                        .errorName(message)
+                        .label(label)
```
Ortega-Dan, line 212: Not sure if this was coordinated with the FrontEnd, but changing this from error message to error name seems to be not appropriate since calles to this method, like this one: https://github.com/Nuvalence/dsgov-work-manager/blob/181751787463f63e2643b1e072088112f6a108c2/service/src/main/java/io/nuvalence/workmanager/service/service/FormConfigurationService.java#L188-L192 do actually provide a message for the error instead of a name.

service/src/main/java/io/nuvalence/workmanager/service/mapper/FormConfigurationMapper.java
```diff
@@ -340,14 +339,21 @@ private void getAllKeysAndComponentsRecursively(
...
+            }
+        }
         throw new NuvalenceFormioValidationException(
                 NuvalenceFormioValidationExMessage.builder()
                         .formioValidationErrors(
                                 List.of(
                                         NuvalenceFormioValidationExItem.builder()
                                                 .controlName(key)
-                                                .errorMessage(message)
+                                                .errorName(message)
```
Ortega-Dan, line 355: same here about error name.

service/src/main/java/io/nuvalence/workmanager/service/config/exceptions/model/NuvalenceFormioValidationExItem.java
```diff
@@ -11,6 +11,5 @@
 public class NuvalenceFormioValidationExItem {
     private String controlName;
     private String errorName;
-    private String errorMessage;
-    private String formStepKey;
+    private String label;
```
Ortega-Dan, line 14: We could use error name in places where we need to use it, but seems like there are places were error message is appropriate (as a longer and more detailed description string). Im ok with adding label.

service/src/main/java/io/nuvalence/workmanager/service/utils/formconfig/formio/NuvalenceFormioValidator.java
```diff
@@ -204,10 +204,19 @@ private static void addValidationError(
...
             String errorName,
             List<NuvalenceFormioValidationExItem> formioValidationErrors) {
+        String label = "";
+        if (controlName != null) {
+            String[] controlNameParts = controlName.split("\\.");
+            if (controlNameParts.length > 0) {
+                label = controlNameParts[controlNameParts.length - 1];
+            }
+        }
+
```
Ortega-Dan, line 214: since this is done in 2 different places, and since this is a util class, should this be modularized as a public method including the subsequent call to transformCameCaseToSeparateWords method ?

Ortega-Dan (commented):
Minimal changes may be needed. I approve on the approach to achieve the validation and the submit of data.

Ortega-Dan (changes requested):
I approve on the approach to achieve the validation and the submit of data. But after a post-review conversation with Juan, those comments are confirmed as needed.

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
jcsuarezm: Thanks @Ortega-Dan, this has been addressed by calling the method directly.

service/src/main/java/io/nuvalence/workmanager/service/service/FormConfigurationService.java
jcsuarezm: Thanks Dan, the message property has been added back. The front end feature related to this ticket mainly uses the name and calculates the label and id itself.

service/src/main/java/io/nuvalence/workmanager/service/mapper/FormConfigurationMapper.java
jcsuarezm: Thanks, this has been corrected.

service/src/main/java/io/nuvalence/workmanager/service/config/exceptions/model/NuvalenceFormioValidationExItem.java
jcsuarezm: Thanks Dan, this class was left as it was before.

service/src/main/java/io/nuvalence/workmanager/service/utils/formconfig/formio/NuvalenceFormioValidator.java
jcsuarezm: This method was removed as now labels are being calculated in the front end.

service/src/main/java/io/nuvalence/workmanager/service/mapper/FormConfigurationMapper.java
```diff
@@ -340,7 +339,13 @@ private void getAllKeysAndComponentsRecursively(
...
      */
     private void throwFormIoException(String key, String message) {
-
+        String label = "";
+        if (key != null) {
+            String[] controlNameParts = key.split("\\.");
+            if (controlNameParts.length > 0) {
+                label = controlNameParts[controlNameParts.length - 1];
+            }
+        }
```
Ortega-Dan, line 348: This seems to be no longer in use but was not removed.

Ortega-Dan (changes requested):
Only one comment pending on not removed code.

service/src/main/java/io/nuvalence/workmanager/service/mapper/FormConfigurationMapper.java
jcsuarezm: Thanks @Ortega-Dan, this has been removed.

Ortega-Dan (approved):
👍 :shipit:

jgnuvalence (approved):

PR #203, opened 2023-08-22
Modified priority change audit event

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


asultan001:
@jcsuarezm , I don't know that the right approach is to record the labels as part of the new and old values since those can technically change over time but the values wouldn't change. Instead I believe that this may be better handled in the front end like we do for other values and resolve the labels at runtime. 

@JayStGelais what do you think?

JayStGelais (approved):

PR #209, opened 2023-08-25
DSG-1563/ignore-computed-attribues-in-audit-events

afcalderonm:
## Description
change the code to remove the computedFields from the audit events

## How Has This Been Tested?
tested locally creating a new application from the front end and verifying from the agency portal that the activity logs does  not show the computed Fields

image

## Types of changes

Patch
Minor
Major

## Checklist:


asultan001 (changes requested):
@afcalderonm this can't be implemented like this. This needs to be dynamic based on the type of field it is. We can't hard code this to look for certain keys.

service/src/main/java/io/nuvalence/workmanager/service/audit/transaction/DynamicDataChangedAuditHandler.java
```diff
@@ -81,4 +86,19 @@ public void publishAuditEvent(OffsetDateTime eventTime, String originatorId) {
             log.error(errorMessage, e);
         }
     }
+
+    private void removeComputedFields(
```
asultan001, line 90: @afcalderonm this seems like it would work but we need tests to cover all of these scenarios.

afcalderonm: @asultan001  i just add the test

JayStGelais (approved):

jgnuvalence (approved):

jgnuvalence (approved):

asultan001 (approved):
LGTM

PR #212, opened 2023-08-30
DSG-1565/fix-timestmap-audit-event-when-updating-documents

afcalderonm:
image

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
```diff
@@ -815,7 +815,7 @@ public AuditEventId postAuditEventForDocumentStatusChanged(
                 document.getCreatedBy(),
                 document.getCreatedBy(),
                 summary,
-                document.getLastUpdatedTimestamp(),
+                document.getReviewedOn(),
```
JayStGelais, line 818: This must use `RequestContextTimestamp.getCurrentTimestamp()`.

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java
afcalderonm: before sending the event the ReviewedOn is set with `RequestContextTimestamp.getCurrentTimestamp()` should i call here too?

JayStGelais: Yes, lets always use the value from `RequestContextTimestamp` for all audit events. If we ever change the logic for updating `reviewedOn` in the future we wont break this timestamp.

asultan001: Wait, I think that is a potential disconnect. Ideally we want to show the same times when you are looking the the audit event and the last reviewed on time so there may be value in using the exact same time and not something that may be off by a few milliseconds.

JayStGelais: The version that first appeared in this PR follows this flow:

`RequestContextTimestamp.getCurrentTimestamp()` -> `reviewedOn` -> Event Timestamp

I'm proposing we switch to:

`RequestContextTimestamp.getCurrentTimestamp()` -> `reviewedOn` 
`RequestContextTimestamp.getCurrentTimestamp()` ->  Event Timestamp

They will still have the exact same timestamp, the one being held in `RequestContextTimestamp`. But now if we move the update to `reviewedOn`  into a JPA lifecycle hook, like we have done for similar timestamps in other entities, we wont introduce a bug where the Event Timestamp is based on old data.

afcalderonm: i just change it to use the `RequestContextTimestamp.getCurrentTimestamp() `

asultan001 (approved):
LGTM

JayStGelais (approved):

PR #254, opened 2023-09-27
DSG-1663/DSG-1751/schemaChangeTrackingAttributes

Ortega-Dan:
Ticket: https://dol-ewf.atlassian.net/browse/DSG-1751

## Description

Schema change tracking attributes automatically managed.

## Motivation and Context

Needed for Schemas FE interface

## How Has This Been Tested?

API and automated tests.

image

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/test/java/io/nuvalence/workmanager/service/domain/dynamicschema/SchemaTest.java
```diff
@@ -37,6 +37,8 @@ void equalsHashcodeContract() {
                                                 new DynaProperty("foo", String.class),
                                                 new DynaProperty("baz", String.class))
                                         .toArray(new DynaProperty[0])))
+                .withIgnoredFields(
+                        "createdBy", "lastUpdatedBy", "createdTimestamp", "lastUpdatedTimestamp")
```
Ortega-Dan, line 41: This tests both that the field exists and that those are not to be considered for equals or hashcode.

service/src/main/java/io/nuvalence/workmanager/service/domain/dynamicschema/Schema.java
```diff
@@ -36,6 +38,12 @@ public final class Schema implements DynaClass {
 
     @Getter private final Map<String, List<AttributeConfiguration>> attributeConfigurations;
 
+    // change tracking
+    @Getter @Setter private String createdBy;
+    @Getter @Setter private String lastUpdatedBy;
+    @Getter @Setter private OffsetDateTime createdTimestamp;
+    @Getter @Setter private OffsetDateTime lastUpdatedTimestamp;
```
Ortega-Dan, line 45: This fields should not be used for hashcode or equals, these fields are read only in practice but should also be independent of a schemaJson object. This class is different to others, because for Schemas we have this middle layer object to communicate between SchemaModel and SchemaRow, which is different to other objects that map directly between entity and model.

jcsuarezm (approved):

PR #276, opened 2023-10-19
DSG-2043/create-default-form-config-for-new-trasanction-definition

afcalderonm:
## Description
Create default fomr configuration for new transaction definitions

## How Has This Been Tested?
Modify unit testing to test the new behavior also tested locally using postman

## Types of changes

Patch
Minor
Major

## Checklist:


asultan001:
@afcalderonm , I'm not sure if this is what we needed. My thought is that during the creation process, we would just create an empty form configuration and set it as the default. If I understand your implementation correctly, you check to see if there are any defaults from before and if not, you create a new one and set it but it seems like you are always using a particular name?

service/src/test/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImplTest.java
```diff
@@ -448,7 +452,11 @@ void putTransactionDefinition() throws Exception {
                                 .content(postBody)
                                 .contentType(MediaType.APPLICATION_JSON))
                 .andExpect(status().isOk())
-                .andExpect(jsonPath("$.name").value("test transaction"));
+                .andExpect(jsonPath("$.name").value("test transaction"))
+                .andExpect(jsonPath("$.defaultFormConfigurationKey").value("testDefault"));
+
+        verify(formConfigurationService, times(1)).getFormConfigurationByKeys(any(), any());
+        verify(formConfigurationService, times(1)).saveFormConfiguration(any());
```
asultan001, line 459: Can we check for the expectation instead of checking that the methods were called? I rather not test the implementation if we can validate the outcome instead.

service/src/test/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImplTest.java
```diff
@@ -472,9 +480,12 @@ void putTransactionDefinitionForbidden() throws Exception {
                                 .content(postBody)
                                 .contentType(MediaType.APPLICATION_JSON))
                 .andExpect(status().isForbidden())
-                .andExpect(jsonPath("$.name").doesNotExist());
+                .andExpect(jsonPath("$.name").doesNotExist())
+                .andExpect(jsonPath("$.defaultFormConfigurationKey").doesNotExist());
 
         verify(transactionDefinitionService, never()).saveTransactionDefinition(any());
+        verify(formConfigurationService, never()).getFormConfigurationByKeys(any(), any());
+        verify(formConfigurationService, never()).saveFormConfiguration(any());
```
asultan001, line 488: Same here

afcalderonm:
> @afcalderonm , I'm not sure if this is what we needed. My thought is that during the creation process, we would just create an empty form configuration and set it as the default. If I understand your implementation correctly, you check to see if there are any defaults from before and if not, you create a new one and set it but it seems like you are always using a particular name?

i made it that way because the endpoint can create or update, in case of the update if the defaultFormConfigurationKey is send, it will just update the transactionDefinition defaultFormConfigurationKey, and also i check if there is already a defaultFormConfiguration created for the given transaction definition otherwise it will attemp to create a new one everytime the transactionDefinition is updated, and the name i made it following the story to be {transactionDefintionKey}Default

service/src/test/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImplTest.java
afcalderonm: it is hard to validate the full outcome because we are not really returning all the form configuration info in this endpoint, that why i validate that we get the define defaultFormConfigurationKey

service/src/test/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImplTest.java
afcalderonm: this one is justo to validate that the methods are not called when there is an error

service/src/test/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImplTest.java
asultan001: Ok, I understand.

service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
```diff
@@ -578,4 +594,18 @@ private PagedWorkflowModel generatePagedWorkflowModel(Page<WorkflowModel> workfl
         model.setPagingMetadata(pagingMetadataMapper.toPagingMetadata(workflows));
         return model;
     }
+
+    private void createDefaultFormConfiguration(TransactionDefinition transactionDefinition) {
+        FormConfigurationUpdateModel defaultFormConfiguration = new FormConfigurationUpdateModel();
+        defaultFormConfiguration.setSchemaKey(transactionDefinition.getSchemaKey());
+        defaultFormConfiguration.setName("Default Form");
+        defaultFormConfiguration.setDescription(
+                "Default for".concat(" ").concat(transactionDefinition.getName()));
```
asultan001, line 603: I think this should say Default form for...

asultan001 (approved):
LGTM

service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
afcalderonm: Fixed

asultan001 (approved):
LGTM

jcsuarezm (approved):

PR #284, opened 2023-10-24
DSG-1913/Upgrade-Java-version

afcalderonm:
## Description
Upgrade java version to 17

## Types of changes

Patch
Minor
Major

## Checklist:


gradlew
```diff
@@ -210,10 +210,10 @@ set -- \
...
 
 # Stop when "xargs" is not available.
-if ! command -v xargs >/dev/null 2>&1
-then
-    die "xargs is not available"
-fi
+##if ! command -v xargs >/dev/null 2>&1
+##then
+##    die "xargs is not available"
+##fi
```
asultan001, line 216: Why is this commented out? should this be removed all together?

service/src/main/resources/application.yml
```diff
@@ -20,6 +20,8 @@ spring:
       write-dates-as-timestamps: false
     deserialization:
       adjust_dates_to_context_time-zone: false
+  config:
+    import: sm://
```
asultan001, line 24: why is this needed?

gradlew
afcalderonm: yeah it needs to be removed because openjdk 17 does not have xargs by default the cloud worker always fails by that validation

service/src/main/java/io/nuvalence/workmanager/service/utils/RequestContextTimestamp.java
```diff
@@ -14,8 +14,8 @@
 public class RequestContextTimestamp {
     private final OffsetDateTime currentTimestamp;
 
-    public RequestContextTimestamp(final Clock requestScopeClock) {
-        this.currentTimestamp = OffsetDateTime.now(requestScopeClock);
+    public RequestContextTimestamp() {
+        this.currentTimestamp = OffsetDateTime.now(Clock.systemUTC());
```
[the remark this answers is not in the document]
afcalderonm: the way the clock was implemented before work for java 11, it was creating a separated Bean base on the java.time.Clock class but in java 17 there was a change that does not allow to access some private, final or non-visible classes, to it was failing to access to some classes of the Clock library and it was failing to create the Bean, i made and  analysis and found out that that bean was just returning the Clock.systemUTC(). so i remove that class and just set the Clock.systemUTC() directly, i talk with Dan who was the person to do that implementation and he says it is OK

gradlew
```diff
@@ -209,12 +209,6 @@ set -- \
         org.gradle.wrapper.GradleWrapperMain \
         "$@"
 
-# Stop when "xargs" is not available.
-if ! command -v xargs >/dev/null 2>&1
-then
-    die "xargs is not available"
-fi
-
```
asultan001, line 217: why was this removed?

afcalderonm: these are  the same lines that was commented in the previous conversation

asultan001: If it is not available in openjdk17 is there an alternative? the next block seems to indicate that we are uxing xargs to parse quoted args and does a bunch of work.

afcalderonm: yeah there is an alternative when we can install from other source before calling the gradlew, but we are not really using any quoted args that needs xargs and it is also removed in the other services, so i dont we really need to do it at least at this moment

JayStGelais: I'm not fond of manually editing the gradlew shell script. This should be generated by the wrapper task and we shouldn't have to touch it.

JayStGelais (approved):

PR #297, opened 2023-11-07
Feature/dsg 2227/filter by set

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -225,9 +227,23 @@ public ResponseEntity<PagedTransactionModel> getTransactions(
...
+        List<String> transactionDefinitionKeysList = new ArrayList<>();
+        if (transactionDefinitionSetKey != null) {
+            transactionDefinitionKeysList =
+                    transactionDefinitionService
+                            .getTransactionDefinitionsBySetKey(transactionDefinitionSetKey)
+                            .stream()
+                            .map(TransactionDefinition::getKey)
+                            .toList();
+
+            if (transactionDefinitionKeysList.isEmpty()) {
```
jcsuarezm, line 239: This condition was required to guard against an edge case in which a transaction definition set key query parameter is provided but there are no transactions associated with it. If this was left to be an empty array, the filters would treat it as if no transaction set key was provided, returning all values regardless of their association to a transaction definition set.

service/src/main/java/io/nuvalence/workmanager/service/models/SearchTransactionsFilters.java
```diff
@@ -83,11 +83,10 @@ public Specification<Transaction> getTransactionSpecifications() {
             Join<TransactionDefinition, Transaction> transactionDefinitionJoin =
                     root.join("transactionDefinition");
 
-            addEqualPredicate(
+            addInPredicate(
```
jcsuarezm, line 86: To support querying transactions by set key, filters were modified to support a list of transaction definition keys instead of a single value. This is important, as sets can have more than one transaction definition associated with them.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jgnuvalence: Is this what we want? normally if we specify query param and there are no results that match it, it simply returns empty, why is it different in this case?

jgnuvalence (approved):
Approved with a comment

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -225,9 +227,23 @@ public ResponseEntity<PagedTransactionModel> getTransactions(
...
+                return ResponseEntity.ok().body(createEmptyPagedTransactionModel(pageSize));
+            }
+        }
+
         TransactionFilters filters =
                 SearchTransactionsFilters.builder()
-                        .transactionDefinitionKey(transactionDefinitionKey)
+                        .transactionDefinitionKey(transactionDefinitionKeysList)
```
Ortega-Dan, line 246: Looks like we are forgetting about the transactionDefinitionKey filter here. The transaction definition set might not be applicable but the transactionDefinitionKey may still apply.

service/src/main/java/io/nuvalence/workmanager/service/models/SearchTransactionsFilters.java
Ortega-Dan: I believe we can use a null list to avoid adding a predicate when the defKey was not provided.

Ortega-Dan: To achieve that looks like the '!attributeValues.isEmpty()' validation needs to be removed in the addInPredicate method that is down this file.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -325,7 +341,7 @@ public ResponseEntity<List<TransactionCountByStatusModel>> getTransactionCountBy
         }
         SearchTransactionsFilters filters =
                 SearchTransactionsFilters.builder()
-                        .transactionDefinitionKey(transactionDefinitionKey)
+                        .transactionDefinitionKey(List.of(transactionDefinitionKey))
```
Ortega-Dan, line 344: Should these counters also work by 'SetKey' instead of my 'DefKey' alone, since these counters are intended for the dashboard ?

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jcsuarezm: Hi Julián, thank you. The behavior you described is what this condition achieves, there is an above condition that checks if there was a transaction set key provided. The service looks for all transaction definition sets that are linked to the set and if none is found it returns an empty array. Without this, the filters below this code would behave as if the set key param was not provided.

jgnuvalence: thank you Juan I misunderstood this the first time, resolving the comment

service/src/main/java/io/nuvalence/workmanager/service/models/SearchTransactionsFilters.java
jcsuarezm: Thank you Dan, a new predicate construction logic has been implemented for this.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jcsuarezm: Thank you Dan, logic at service level was added to correct this.

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
jcsuarezm: Logic was added to query by set in this endpoint.

service/src/main/java/io/nuvalence/workmanager/service/models/SearchTransactionsFilters.java
```diff
@@ -47,7 +47,7 @@ public class SearchTransactionsFilters extends TransactionFilters {
      */
     @Builder
     public SearchTransactionsFilters(
-            String transactionDefinitionKey,
+            List<String> transactionDefinitionKey,
```
Ortega-Dan, line 50: ultra last comment. Better to have this variable in plural to have it clear, otherwise you can get confused for a little when reading the code.

Ortega-Dan (approved):
:+1:

service/src/main/java/io/nuvalence/workmanager/service/models/SearchTransactionsFilters.java
jcsuarezm: Thank you, the variable has been renamed

Ortega-Dan (approved):
:shipit:

PR #300, opened 2023-11-08
Corrected schema response fields

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
```diff
@@ -190,10 +191,14 @@ public ResponseEntity<SchemaModel> updateSchema(
...
         schemaId = existingSchema.get().getId();
+        schemaService.saveSchema(
+                dynamicSchemaMapper.schemaUpdateModelToSchema(schemaUpdateModel, key, schemaId));
+
         Schema schema =
-                schemaService.saveSchema(
-                        dynamicSchemaMapper.schemaUpdateModelToSchema(
-                                schemaUpdateModel, key, schemaId));
+                schemaService
+                        .getSchemaByKey(key)
```
jcsuarezm, line 199: An additional call to get the resulting schema was added. The service level method for saving is transactional and does several mappings between the saved SchemaRow object back to the Schema that is returned. Tracking attributes are only introduced with PrePersist and PreUpdate hooks, which due to the service's method transactional nature, are only executed after mapping and return, causing the bug behavior. It was opted to do this additional call instead of removing the transactional annotation, due to the possibility of regression or bug introduction as a result. As schemas are seldom updated or created, this is not expected to represent a significant impact on performance.

service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
```diff
@@ -190,10 +191,14 @@ public ResponseEntity<SchemaModel> updateSchema(
...
+                        .orElseThrow(
+                                () -> new UnexpectedException("Schema not found after saving"));
```
jcsuarezm, line 201: This exception is never expected, as the object was just saved. It was included to prevent SonarQube bugs.

service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
```diff
@@ -210,9 +215,13 @@ public ResponseEntity<SchemaModel> createSchema(SchemaCreateModel schemaCreateMo
...
 
+        schemaService.saveSchema(dynamicSchemaMapper.schemaCreateModelToSchema(schemaCreateModel));
+
         Schema schema =
-                schemaService.saveSchema(
-                        dynamicSchemaMapper.schemaCreateModelToSchema(schemaCreateModel));
+                schemaService
+                        .getSchemaByKey(schemaCreateModel.getKey())
+                        .orElseThrow(
+                                () -> new UnexpectedException("Schema not found after saving"));
```
Ortega-Dan, line 224: Seems like this adjustment is not really needed. Since it is already working well for create transaction. The following is a call to the current dev env.

We can also keep it if we just want to have consistent code in both calls:

Ortega-Dan (commented):
Looks like the PR title here also needs to be adjusted to reflect the final aproach implemented (or just the fix achieved)

Ortega-Dan (approved):

service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java
jcsuarezm: Thank you Dan, I have removed unnecessary code, to prevent unneeded complexity.

Ortega-Dan (approved):

PR #324, opened 2023-11-27
checking actuator behavior

Ortega-Dan:
App is failing:
image
and rollout fails:
image
instead of 
image
but looks like we are not easily able to check for rollout success for now. 
So this change is just a test related to actuator.


Ortega-Dan:
after our conversation with @aardvarq . I realize this may be a valid test. But this is likely to be checked as part of a specific ticket related work. So this may stay here as a closed PR for reference, or reopening if needed.

PR #340, opened 2023-12-07
DSG-2343/create-Record-Domain-and-API

afcalderonm:
## Types of changes

Patch
Minor
Major

## Checklist:


swagger.yaml
```diff
@@ -2510,6 +2634,71 @@ components:
...
+      properties:
+        data:
+          type: object
+          additionalProperties: true
+        status:
+          type: string
+        expires:
+          type: string
+          format: date-time
+    RecordModel:
```
Ortega-Dan, line 2661: I would recommend using RecordResponseModel instead, it makes it more clear but this is nit, I know we don't have consistency on this. But my recommendation.

service/src/main/java/io/nuvalence/workmanager/service/mapper/RecordMapper.java
```diff
@@ -0,0 +1,27 @@
...
+import io.nuvalence.workmanager.service.generated.models.RecordModel;
+import lombok.Setter;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.springframework.beans.factory.annotation.Autowired;
+
+@Mapper(componentModel = "spring")
+public abstract class RecordMapper {
+
+    @Autowired @Setter protected EntityMapper entityMapper;
```
Ortega-Dan, line 13: Constructor injection is recommended for this because this is required. And having a constructor makes it clear it is. We can use @AllArgsConstructor

helm/work-manager/cerbos-policies/record_resource.yaml
```diff
@@ -0,0 +1,16 @@
...
+      roles:
+        - transaction-submitter
+        - transaction-admin
+        - transaction-config-admin
+      effect: EFFECT_ALLOW
+
+    - actions: ["create", "update", "delete"]
+      roles:
+        - transaction-config-admin
+      effect: EFFECT_ALLOW
```
Ortega-Dan, line 16: Do we know if the record subject must also be able to update their records ?
This is just a question:
ref: https://github.com/Nuvalence/dsgov-work-manager/blob/36d733b58ea0f7908f72ea06e58e2196d10520bf/helm/work-manager/cerbos-policies/transaction_resource.yaml#L18-L18

afcalderonm: Yeah records should work with the same permissions as transactions, it was change from the las time i check it, i just made the changes thank you dan!

swagger.yaml
afcalderonm: yeah you right, i just change to RecordResponseModel

service/src/main/java/io/nuvalence/workmanager/service/mapper/RecordMapper.java
afcalderonm: done

service/src/main/java/io/nuvalence/workmanager/service/service/RecordService.java
```diff
@@ -0,0 +1,89 @@
...
+        return repository.findAll(filter.getRecordSpecification(), filter.getPageRequest());
+    }
+
+    public Record updateRecord(RecordUpdateRequest updateRequest, Record existingRecord) {
+
+        if (validateBlankOrEmpty(updateRequest.getStatus())) {
+            existingRecord.setStatus(updateRequest.getStatus());
+        }
+
+        if (validateBlankOrEmpty(String.valueOf(updateRequest.getExpires()))) {
```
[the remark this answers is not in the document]
afcalderonm: thank you dan i just change it to use the StringUtils.isBlank() method.

service/src/main/java/io/nuvalence/workmanager/service/service/RecordService.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public Record updateRecord(RecordUpdateRequest updateRequest, Record existingRecord) {
+
+        if (!StringUtils.isBlank(updateRequest.getStatus())) {
+            existingRecord.setStatus(updateRequest.getStatus());
+        }
+
+        if (!StringUtils.isBlank(String.valueOf(updateRequest.getExpires()))) {
+            existingRecord.setExpires(updateRequest.getExpires());
```
Ortega-Dan, line 65: Andres, if subject is able to update the record, they must not be able to update the status or the expiration. That is an protection needed for that data. I would recommend asking Jay if we want to have this protection from the beginning or not, but it is anyway needed.
In case it helps. It may not be a hard lift and transactions also have that protection:
refs:
1) https://github.com/Nuvalence/dsgov-work-manager/blob/24e87b2bdf7ecf7416558822de4bcd272ce22030/service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java#L450-L459
2) https://github.com/Nuvalence/dsgov-work-manager/blob/24e87b2bdf7ecf7416558822de4bcd272ce22030/service/src/main/java/io/nuvalence/workmanager/service/service/TransactionService.java#L931-L943

service/src/main/java/io/nuvalence/workmanager/service/mapper/RecordMapper.java
Ortega-Dan: Sorry for not being more explicit if we have the proper pattern with constructor we should remove the @Autowired

service/src/main/java/io/nuvalence/workmanager/service/service/RecordService.java
afcalderonm: i just add the protection

service/src/main/java/io/nuvalence/workmanager/service/mapper/RecordMapper.java
afcalderonm: removed

Ortega-Dan (approved):
:rocket:

PR #343, opened 2023-12-12
Bug/DSG-2481/Return 200 when no document processors present on process enqueue

kives-nu:
## Description
Updating document processor logic to return 200 when no processors present. 

[DSG-2481](https://dol-ewf.atlassian.net/browse/DSG-2481)
[Related PR for DM](https://github.com/Nuvalence/dsgov-document-management/pull/132)

## Motivation and Context
When no document processors are attached to a document, the product guidance is to return 200 when we attempt to process, instead of an error code. Prior to these changes, a 403 would be thrown as the schema parsing function would through a null pointer error instead of returning a list of empty processors. This PR addresses that null pointer exception and adds a test to verify it. There is a corresponding document manager PR as well, to return 200 instead of 400 when it receives no processors. 

## Areas for Focus
I also made a change to the values.local.yml in the helm charts. I am not sure if this should be kept or not. It appears to me like it was an error, dsgov-document-manager does not align with the other values in the local.yml. This particular line broke the ability of my local minikube work manager to communicate my minikube document manager. Since the default name for the local service is document-manager, without the dsgov. 

However, the name has been dsgov-document-manager for 7 months. I would have thought this breaks all communicate between local WM and local DM for everyone, but I am assuming it is not given how long the change has been in. If anyone has any insight into why and if the values.local.yml should be kept the way it is, that would be great. 

## How Has This Been Tested?
If you follow the steps to reproduce on the ticket (you may need to modify a schema to have no document processors first), you will see 403 errors. If you deploy this fix (and the DM fix) and reproduce, the document processing endpoint will return 200 with an empty list instead. Additionally, there is a new unit test to verify the new null check. 

## Types of changes
Patch
Minor
Major

## Checklist:

[DSG-2481]: https://dol-ewf.atlassian.net/browse/DSG-2481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helm/work-manager/values.local.yaml
```diff
@@ -66,7 +66,7 @@ env:
   SELF_SIGN_PUBLIC_KEY: "${sm://projects/dsgov-dev/secrets/dsgov-self-signed-token-public-key/versions/latest}"
   SELF_SIGN_PRIVATE_KEY: "${sm://projects/dsgov-dev/secrets/dsgov-self-signed-token-private-key/versions/latest}"
   USER_MANAGEMENT_BASE_URL: "http://user-management/um"
-  DOCUMENT_MANAGEMENT_BASE_URL: "http://dsgov-document-management/dm"
+  DOCUMENT_MANAGEMENT_BASE_URL: "http://document-management/dm"
```
kives-nu, line 69: See PR description for discussion on this change

jgnuvalence: Personally, this works well for me with the dsgov prefix, in my local I can see that docman has that prefix in the service release name, but the other apps do not have it, I think the fix should be done there before changing it here, but if this it works for you gets me confused

jgnuvalence: btw now we are communicating with docman eventually, so anyway this is not going to be used, we probably need a ticket to remove this API client

jgnuvalence (approved):

helm/work-manager/values.local.yaml
kives-nu: Makes sense, I can revert the change since then. 

I am running my DM service through cloudclode and attaching a debugger, not sure if that's somehow impacting the service name

jgnuvalence (changes requested):
reviewing the docman PR I realize that probably the best approach is to simply not send the event to start document processing when there is no processor, this way you don't need to make any changes to document-management

service/src/main/java/io/nuvalence/workmanager/service/controllers/TransactionsApiDelegateImpl.java
```diff
@@ -301,6 +301,14 @@ public ResponseEntity<InitiateDocumentProcessingModelResponse> initiateDocumentP
         List<String> processorsNames =
                 schemaService.getDocumentProcessorsInASchemaPath(request.getPath(), schema);
 
+        if (processorsNames.isEmpty()) {
```
jgnuvalence, line 304: I think we still need to save the customer-provided documents even if they don't have processors, I think we just need to avoid executing:
documentManagementService.initiateDocumentProcessing(...)

jgnuvalence (approved):

PR #349, opened 2023-12-14
Feature/dsg 2494/employer profiles

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/domain/profile/Employer.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Column(name = "fein", nullable = false)
+    private String fein;
+
+    @Column(name = "legal_name", nullable = false)
+    private String legalName;
+
+    @ElementCollection(targetClass = String.class, fetch = FetchType.EAGER)
+    @CollectionTable(name = "other_names", joinColumns = @JoinColumn(name = "employer_profile_id"))
+    @Column(name = "other_names", nullable = false)
```
jgnuvalence, line 57: This is maybe not that important, I just think this name is a bit confusing, since this is specific to employer profiles, maybe employer_profile_other_names would be better? Take it as a question, if you don't agree that's fine

service/src/main/java/io/nuvalence/workmanager/service/controllers/EmployerProfilesApiDelegateImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+        Employer employerToBeSaved =
+                employerMapper.updateModelToEmployer(employerProfileUpdateModel);
+        employerToBeSaved.setId(profileId);
+        employerToBeSaved.setCreatedBy(existingEmployer.getCreatedBy());
+        employerToBeSaved.setCreatedTimestamp(existingEmployer.getCreatedTimestamp());
+
+        employerService.saveEmployer(employerToBeSaved);
+
+        Employer savedEmployer =
```
jgnuvalence, line 122: Is this necessary? We already know that the emproyerProfile exists from the query above, couldn't you just return:

Employer employer = employerService.saveEmployer(...)

?

service/src/main/resources/db/changelog/changelog/db.changelog-030.xml
```diff
@@ -0,0 +1,95 @@
...
+            </column>
+            <column name="employer_for_mailing_id" type="uuid">
+                <constraints nullable="true" foreignKeyName="fk_address_mailing_employer"
+                             references="employer_profile(id)"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet id="create_other_names_table" author="jcsuarezm">
+        <createTable tableName="other_names">
```
jgnuvalence, line 85: This is maybe not that important, I just think this name is a bit confusing, since this is specific to employer profiles, maybe employer_profile_other_names would be better? Take it as a question, if you don't agree that's fine

jgnuvalence (approved):

service/src/main/resources/db/changelog/changelog/db.changelog-030.xml
jcsuarezm: Thank you Julián, the table name has been changed to employer_profile_other_names  for clarity while keeping the attribute name to just other_names for simplicity.

service/src/main/java/io/nuvalence/workmanager/service/controllers/EmployerProfilesApiDelegateImpl.java
jcsuarezm: Hi Julián, thank you. I believe it is necessary to have this call after the save. The service class marked as Transactional has caused fields such as last updated time to not be present or be incorrectly set after just the save, as these are all set and applied by the listener, which runs on PreUpdate. Similar approaches as this one were used to correct this issue, as seen here: https://github.com/Nuvalence/dsgov-work-manager/blob/main/service/src/main/java/io/nuvalence/workmanager/service/controllers/AdminApiDelegateImpl.java#L239. 

Please let me know if you agree.

jgnuvalence: ok makes sense, thank you!

jgnuvalence (approved):

PR #354, opened 2023-12-14
Dsg 1421/publish audit events through dsgov-events

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


settings.gradle
```diff
@@ -12,6 +12,5 @@ rootProject.name = 'work-manager'
 include 'service'
 include 'client'
 include 'acceptance-tests'
-include 'audit-service-client'
 include 'notification-service-client'
```
jcsuarezm, line 15: I think this should be deleted too.

config/checkstyle/checkstyle.xml
```diff
@@ -44,7 +44,7 @@
 
     <module name="LineLength">
         <property name="fileExtensions" value="java"/>
-        <property name="max" value="100"/>
+        <property name="max" value="120"/>
```
jcsuarezm, line 47: I agree that extending the line length could be a good idea too. I might be wrong but I think Google's style is set to 100, but overall 120 seems like a better fit for our project, but just wanted to point this out.

jcsuarezm (approved):

settings.gradle
jgnuvalence: you are right, I'll delete them

config/checkstyle/checkstyle.xml
jgnuvalence: I see several parts in our project with lines longer than 100 chars, I think that if we want to modify these things we should create a ticker to fix this in a general way, for now I will leave it like this, do you agree?

jcsuarezm (approved):

PR #366, opened 2023-12-19
Feature/dsg 2497/employer user links

kives-nu:
## Description
Adds CRUD for the employer user links [as defined here](https://nuvalence.atlassian.net/wiki/spaces/DSGOV/pages/1637122072/Employer+and+Individual+Profiles#Employer-Profile-API) 

## Motivation and Context
This adds the ability to link employer profiles to specific users, retrieve them, and delete those links. 

## How Has This Been Tested?
Unit tests have been added for the controller, service, mapper and entity. To Dev test, you need a user in UM service, and UM service needs to be running. You then create an employer profile, and use the PUT endpoint from this PR to link a user to that profile. You can then test the get endpoint and its filtering options, and finally you can delete the link and see that you cannot retrieve the link anymore with the get request. 

## Types of changes

Patch
Minor
Major

## Checklist:
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


service/src/main/java/io/nuvalence/workmanager/service/models/IndividualFilters.java
```diff
@@ -3,15 +3,13 @@
 import io.nuvalence.workmanager.service.domain.profile.Individual;
```
kives-nu, line 3: Changes to this file were necessary because I modified the behavior of the userManagementClient function

service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
```diff
@@ -56,35 +61,53 @@ public Optional<User> getUser(UUID userId) {
     }
 
     @Override
```
kives-nu, line 63: The primary motivation for these changes in particular was to change the pagination behavior of the function to get all relevant pages so no users were left behind. It was also easier to return a list of userDTOs object that wasn't wrapped in a page.

swagger.yaml
```diff
@@ -2076,6 +2178,45 @@ components:
...
+        profileId:
+          type: string
+          format: uuid
+        userId:
+          type: string
+          format: uuid
+        profileAccessLevel:
+          type: string
+          pattern: ^(READER|WRITER|ADMIN|AGENCY_READONLY)
+        createdBy:
```
Ortega-Dan, line 2199: It is recommended to have the response bring all the properties from the request like in: https://github.com/Nuvalence/dsgov-work-manager/blob/bdba636ada877e7ec816f6cad4d3a574ad0270a8/swagger.yaml#L2150-L2167

swagger.yaml
```diff
@@ -2076,6 +2178,45 @@ components:
             $ref: "#/components/schemas/EmployerProfileResponseModel"
         pagingMetadata:
           $ref: "#/components/schemas/PagingMetadata"
+    EmployerProfileLinkRequest:
```
Ortega-Dan, line 2181: We also tend to call this ones ...Model to clearly differentiate from the internal objects.

service/src/main/java/io/nuvalence/workmanager/service/usermanagementapi/UserManagementClient.java
Ortega-Dan: :+1: Nice.

service/src/main/java/io/nuvalence/workmanager/service/controllers/EmployerProfilesApiDelegateImpl.java
```diff
@@ -81,6 +91,50 @@ public ResponseEntity<PageEmployerProfileResponseModel> getEmployerProfiles(
...
+
+        List<UUID> userIds;
+        if ((!StringUtils.isBlank(name) || !StringUtils.isBlank(email)) && userId == null) {
+            List<UserDTO> users = userManagementService.getUsers(name, email);
+            userIds = users.stream().map(UserDTO::getId).toList();
+        } else if (userId != null) {
+            userIds = new ArrayList<>(List.of(userId));
+        } else {
+            userIds = null;
+        }
```
Ortega-Dan, line 117: I just believe we do not really need this final else, since it is already null in line 109.

service/src/main/resources/db/changelog/changelog/db.changelog-034.xml
```diff
@@ -0,0 +1,38 @@
...
+    <changeSet id="create_employer_user_link_table" author="kives">
+        <createTable tableName="employer_user_link">
+            <column name="id" type="uuid">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="profile" type="uuid">
+                <constraints nullable="false" foreignKeyName="fk_employer_user_link_employer"
+                             references="employer_profile(id)"/>
+            </column>
+            <column name="user_id" type="varchar(255)">
```
Ortega-Dan, line 16: is there a reason why this is not uuid ?

service/src/main/java/io/nuvalence/workmanager/service/service/EmployerUserLinkService.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.UUID;
+
+import jakarta.transaction.Transactional;
+
+@Service
+@Transactional
+@RequiredArgsConstructor
+public class EmployerUserLinkService {
+    private final EmployerUserLinkRepository repository;
+    private final UserManagementClient userManagementClient;
```
Ortega-Dan, line 21: nit: looks like this line is not really needed here.

service/src/main/java/io/nuvalence/workmanager/service/models/EmployerUserLinkFilters.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.Getter;
+import org.springframework.data.jpa.domain.Specification;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import jakarta.persistence.criteria.Predicate;
+
+@Getter
```
Ortega-Dan, line 14: about the errors in the pipeline, it is possible it is due to several SonarQube warnings. like kissing javadocs in this class. This class has several warnings and you can tackle them locally if you install a pluging called SonarLint (not sure if it is the same name in intellij but I use that one in vscode) which may help to identify warnings. 
I also know some pre-existing files may have the same issues. But better if we see if this is related since warnings have special weight in new code.

service/src/main/java/io/nuvalence/workmanager/service/domain/profile/EmployerUserLink.java
```diff
@@ -0,0 +1,73 @@
...
+    @Column(name = "created_timestamp", nullable = false)
+    protected OffsetDateTime createdTimestamp;
+
+    @Column(name = "last_updated_timestamp", nullable = false)
+    private OffsetDateTime lastUpdatedTimestamp;
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof EmployerUserLink link)) return false;
```
Ortega-Dan, line 63: here are some checkstyle errors, those should pop if you run locally: `./gradlew clean build`

Ortega-Dan: You could also use a checkstyle plugin. For vscode I use: https://marketplace.visualstudio.com/items?itemName=rianadon.vscode-java-checkstyle, but that is just a helper, running the gradlew clean build should be enough since it produces a report. and the output will have a link to the local html report.

service/src/test/java/io/nuvalence/workmanager/service/models/IndividualFiltersTest.java
```diff
@@ -146,10 +146,8 @@ void testIndividualSpecificationWithEmail() {
                         null,
                         userManagementService);
 
-        UserPageDTO userPageDTO = new UserPageDTO();
-        userPageDTO.setUsers(List.of(createUserDto()));
         when(userManagementService.getUsers(any(), eq("email@email.com")))
-                .thenReturn(Optional.of(userPageDTO));
+                .thenReturn(List.of(createUserDto()));
```
Ortega-Dan, line 151: This test file is broken in several tests due to the need to adjust this.

Ortega-Dan (changes requested):
Some comments

swagger.yaml
kives-nu: Do I not have these underneath ? One line 2199 - 2210 ?

service/src/main/resources/db/changelog/changelog/db.changelog-034.xml
kives-nu: I had issues with hibernate not passing UUID objects as such, but converting them to strings. So if this was UUID, all my repository methods were failing. 

I figured we must have seen this problem before so I dug around the transaction entity and found this 

```
    @Setter
    @Column(name = "transaction_definition_id", length = 36, nullable = false)
    @JdbcTypeCode(SqlTypes.VARCHAR)
    private UUID transactionDefinitionId;
 ``` 
 
 Where we are treating a UUID as a string at the DB layer, presumably because of the same issue I was encountering.

swagger.yaml
kives-nu: I renamed to EmployerProfileLinkRequestModel, is that what you meant ?

service/src/main/resources/db/changelog/changelog/db.changelog-034.xml
Ortega-Dan: Ok. Thanks :+1:

swagger.yaml
Ortega-Dan: Right. Thanks. As mentioned, this is more just to keep the pattern and to the actual differentiation.

swagger.yaml
Ortega-Dan: This is more about the usage of allOf to bring properties from another object and then extend adding new ones.

service/src/test/java/io/nuvalence/workmanager/service/models/IndividualFiltersTest.java
Ortega-Dan: Solved

Ortega-Dan (approved):

Ortega-Dan:
Approved but it is not compiling in my local:

PR #373, opened 2024-01-03
DSG-2550/Capture-Profile-User-audit-events

afcalderonm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/audit/profile/EmployerProfileUserAccessLevelChangedAuditHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.UUID;
+
+@Slf4j
+@RequiredArgsConstructor
+public class EmployerProfileUserAccessLevelChangedAuditHandler
+        implements AuditHandler<EmployerUserLink> {
+
+    private String before;
```
JayStGelais, line 18: For readability I would recommend making the before/after attributes clear that they are for storing access level. This can be inferred from context, but its less cognitive effort if we name them more explicitly. 

I'd suggest `accessLevelBefore` and `accessLevelAfter`.

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/EmployerProfileUserAccessLevelChangedAuditHandler.java
```diff
@@ -0,0 +1,69 @@
...
+                eventSummary =
+                        String.format(
+                                "Profile user access level changed to [%s] for employer profile"
+                                        + " user %s owned by %s. Previously it was [%s]",
+                                after, userId, ownerUserId, before);
+            } else {
+                return;
+            }
+
+            individualAuditEventService.sendStateChangeEvent(
```
JayStGelais, line 52: We should probably be including ownerUserId and userId in the data JSON so the front end can parse them out and look up user info when rendering displays.

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/IndividualProfileUserAccessLevelChangedAuditHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import io.nuvalence.workmanager.service.models.auditevents.AuditEventBusinessObject;
+import io.nuvalence.workmanager.service.service.AuditEventService;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.UUID;
+
+@Slf4j
+@RequiredArgsConstructor
+public class IndividualProfileUserAccessLevelChangedAuditHandler
```
JayStGelais, line 15: Same feedback as left for `EmployerProfileUserAccessLevelChangedAuditHandler`

lets use more descriptive names for before and after and include additional fields in the data JSON.

JayStGelais (changes requested):

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/EmployerProfileUserAccessLevelChangedAuditHandler.java
afcalderonm: thanks jay, the change was made

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/IndividualProfileUserAccessLevelChangedAuditHandler.java
afcalderonm: thanks jay, the change was made

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/EmployerProfileUserAccessLevelChangedAuditHandler.java
afcalderonm: thanks jay, the change was made

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/EmployerProfileUserAccessLevelChangedAuditHandler.java
```diff
@@ -27,37 +31,46 @@ public class EmployerProfileUserAccessLevelChangedAuditHandler
...
-                                after, userId, ownerUserId, before);
+                                accessLevelAfter.get(ACCESS_LEVEL),
+                                userId,
+                                ownerUserId,
+                                accessLevelBefore.get(ACCESS_LEVEL));
             } else {
                 return;
             }
 
+            String eventDataJson =
```
JayStGelais, line 59: While this is likely safe because we are only interpolating 2 UUIDs into the string, this pattern would not be prefered for other types of data and might make it more complicated to add additional fields in the future. It's important that whatever values we interpolate in the future are properly escaped for JSON. 

A cleaner approach would look like this:

```
Map<String,Object> eventData = Map.of(
    "ownerUserId", ownerUserId,
    "userId", userId
);
String eventDataJson = SpringConfig.getMapper().writeValueAsString(eventData);
```

service/src/main/java/io/nuvalence/workmanager/service/audit/profile/IndividualProfileUserAccessLevelChangedAuditHandler.java
```diff
@@ -27,36 +31,44 @@ public class IndividualProfileUserAccessLevelChangedAuditHandler
...
-                                after, userId, ownerUserId, before);
+                                accessLevelAfter.get(ACCESS_LEVEL),
+                                userId,
+                                ownerUserId,
+                                accessLevelBefore.get(ACCESS_LEVEL));
             } else {
                 return;
             }
 
+            String eventDataJson =
```
JayStGelais, line 59: We should also take care to escape values for JSON here.

JayStGelais (changes requested):

JayStGelais (approved):

PR #387, opened 2024-01-16
DSG-2659/Filter-ProfileAccessLevel-in-enums-endpoint

afcalderonm:
## Description
add a filter to return diferente enum values depending on the user type

## How Has This Been Tested?
test writing unit test for the new use cases and also testes locally using postman

test for get profile access level enum from an public user 
image

test for get profile access level enum from an agency user 

image

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/EnumsApiDelegateImpl.java
```diff
@@ -84,6 +85,11 @@ private List<EnumerationResponseModelInner> buildEnumResponseModel(
                                 ? getRankedEnumModel((ApplicationEnumRanked) appEnum)
                                 : getEnumModel(appEnum);
 
+                if (UserUtility.getAuthenticatedUserType().equalsIgnoreCase("public")
```
Ortega-Dan, line 88: Looks like this if can be integrated into the previous if to avoid getting an enum that is not going to be used, and avoid an additional if.

Ortega-Dan (approved):
Approved with a need I recommend to adjust for clarity. Approach LGTM.

service/src/main/java/io/nuvalence/workmanager/service/controllers/EnumsApiDelegateImpl.java
afcalderonm: Thank you dan, it is done!

Ortega-Dan (approved):
:+1:

PR #388, opened 2024-01-16
DSG-2551 Post Audit Events for Employer Profile Invitations

kives-nu:
## Description
Adding profile invitation audit events. Events are for invite claimed, invite sent and invite deleted 

## Areas for Focus
I think we have a lot of code duplication between different profile types right now. It might be good to think of ways we can abstract out a lot of the common code. 

## How Has This Been Tested?
Unit tests.

I have confirmed audit events are being added via the DB in the audit service locally. 

## Screenshots
These screenshots represent the three new audit events as pulled out of audit service. 

09C1813A-4574-4E52-A7B5-8011C4E77BF6
AD3F3B21-4316-4785-8F49-FC9589BD36AA
5E6456D4-7F66-4B11-947D-3988E4EA786B

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/IndividualProfilesApiDelegateImpl.java
```diff
@@ -519,4 +535,20 @@ private void postAuditEventForIndividualProfileUserRemoved(
             log.error(errorMessage, e);
         }
     }
+
+    private void postAuditEventForProfileInvitation(
+            ProfileInvitation profileInvitation, AuditActivityType auditActivityType, UUID userId) {
+        try {
+            individualService.postAuditEventForIndividualProfileInvites(
+                    profileInvitation, auditActivityType, userId);
+        } catch (Exception e) {
```
kives-nu, line 544: This is failing sonarrqube on test coverage because I am not testing this exception test. I can write a test that executes these lines, but it would be meaningless because we do not have a framework for mocking our injected logger. The log statement is theoretically testable but we'll need to set up a whole test logger framework. Not sure if this is desirable at the moment, or if the code coverage for sonarrqube can be overrulled in this case.

kives-nu: We also have many similar functions that lack the same test coverage. Its just my PR is smaller in comparison so not testing these lines drops me below 80%

service/src/main/java/io/nuvalence/workmanager/service/service/EmployerService.java
```diff
@@ -129,4 +131,31 @@ public void postAuditEventForEmployerProfileUserRemoved(EmployerUserLink employe
                 profileUserInfo.toJson(),
                 AuditActivityType.EMPLOYER_PROFILE_USER_REMOVED);
     }
+
+    public void postAuditEventForEmployerProfileInvites(
```
kives-nu, line 135: I chose to keep this as one function for all invites since they would all be using the same DTO.

service/src/main/java/io/nuvalence/workmanager/service/controllers/IndividualProfilesApiDelegateImpl.java
```diff
@@ -477,6 +488,11 @@ public ResponseEntity<Void> deleteIndividualProfileInvitation(
 
         profileInvitationService.deleteProfileInvitation(invitationId);
 
+        postAuditEventForProfileInvitation(
+                profileInvitation,
+                AuditActivityType.PROFILE_INVITATION_DELETED,
+                UUID.fromString(SecurityContextUtility.getAuthenticatedUserId()));
```
Ortega-Dan, line 494: nit: this userId is used all the way down as string, you may just passed it as a string here.

service/src/main/java/io/nuvalence/workmanager/service/service/IndividualService.java
```diff
@@ -167,4 +169,31 @@ public void postAuditEventForIndividualProfileUserRemoved(
...
+                switch (auditActivityType) {
+                    case PROFILE_INVITATION_SENT -> "Profile Invitation Sent";
+                    case PROFILE_INVITATION_CLAIMED -> "Profile Invitation Claimed";
+                    case PROFILE_INVITATION_DELETED -> "Profile Invitation Deleted";
+                    default -> "Profile Invitation Event";
+                };
+
+        auditEventService.sendActivityAuditEvent(
+                userId.toString(),
+                userId.toString(),
```
Ortega-Dan, line 192: This is more of a comment. I see here the originator and related user are always being set to the user making the request. Which may not be the actual case since the user is somehow the subject related to the invitation, but I know invitations are intended to emails and those emails may not have a user yet. 
Am I understanding this correctly or are we able to use the subject for the userid, and the authenticated user for the originatorId ?

kives-nu: Yes I believe you're correct. The email is being sent to somebody who *may* not yet have a userID

Ortega-Dan: Or I realize we could use the owner user profile Id, for the user, and continue using authenticated user for the originator. Do you see this as possible ?
This is with the intention to make the events accurately descriptive.

kives-nu: hmm, I'm not sure that's possible either. I don't believe Employer profiles have an owning users, for example.

service/src/main/java/io/nuvalence/workmanager/service/service/EmployerService.java
```diff
@@ -129,4 +131,31 @@ public void postAuditEventForEmployerProfileUserRemoved(EmployerUserLink employe
...
+                    case PROFILE_INVITATION_CLAIMED -> "Profile Invitation Claimed";
+                    case PROFILE_INVITATION_DELETED -> "Profile Invitation Deleted";
+                    default -> "Profile Invitation Event";
+                };
+
+        auditEventService.sendActivityAuditEvent(
+                userId.toString(),
+                userId.toString(),
+                summary,
+                profileInvitation.getId(),
```
Ortega-Dan, line 156: Thanks for using the right businessObjectId :+1:

Ortega-Dan (approved):
LGTM, some comments about possiblitities, but anyway approved if the owerID is not reachable for the notification.

Ortega-Dan (approved):

PR #394, opened 2024-01-18
DSG-2658/Filter-list-of-users-for-a-profile-by-user-type

afcalderonm:
## Description
add a filter to avoid public users to retrieve profile links that have AGENCY_READONLY access level

## How Has This Been Tested?
tested locally using postman

test for get individual links from a public user
image

test for get same individual links from a agency user
image

test for get employer links from a public user
image

test for get same individual links from a agency user
image

## Types of changes

Patch
Minor
Major

## Checklist:


Ortega-Dan (commented):
I would recommend adding some tests for the EmployerUserLink, to protect the functionality, even more when tests are going to be somehow the same to the individual link ones.
I am also not sure if we are really testing that the predicate is added or the condition is met, or just avoiding previous tests to break. But happy to check together if I am not seeing it.

afcalderonm:
> I would recommend adding some tests for the EmployerUserLink, to protect the functionality, even more when tests are going to be somehow the same to the individual link ones. I am also not sure if we are really testing that the predicate is added or the condition is met, or just avoiding previous tests to break. But happy to check together if I am not seeing it.

hey dan they were just added to prevent previous test to fail, but i just add test to validate that specific predicate, thank you!

Ortega-Dan (approved):

PR #409, opened 2024-01-25
Initial implementation

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/controllers/ConversationsApiDelegateImpl.java
```diff
@@ -62,18 +69,10 @@ public ResponseEntity<ConversationResponseModel> postConversation(
                 conversationMapper.createModelToConversation(conversationCreateModel);
         Message message = messageMapper.createModelToMessage(conversationCreateModel.getMessage());
         EntityReference entityReference = new EntityReference();
-        List<MessageParticipant> participants =
```
jcsuarezm, line 65: These deletes in the post method are expected to have no impact, as the whole method will be changed to match the new design.

service/src/main/java/io/nuvalence/workmanager/service/controllers/ConversationsApiDelegateImpl.java
```diff
@@ -88,35 +87,58 @@ public ResponseEntity<ConversationResponseModel> postConversation(
...
             String referenceType,
             UUID referenceId,
-            Boolean unReadMessages,
+            UUID xApplicationProfileID,
             String sortBy,
             String sortOrder,
             Integer pageNumber,
             Integer pageSize) {
 
-        if (!authorizationHandler.isAllowed("view", Conversation.class)) {
```
jcsuarezm, line 97: Cerbos evaluation for this endpoint is now not done for Conversation as a class, but rather a check is done for the profile present in headers. As this profile is used downstream to get the referenced entities, this validation should suffice.

Ortega-Dan (approved):

swagger.yaml
```diff
@@ -4095,36 +4108,6 @@ components:
...
-          type: string
-          description: Type of the user
-        profileId:
-          type: string
-          format: uuid
-          description: Unique identifier for the user's profile
-        profileType:
-          type: string
-          description: Type of the profile
     PageConversationResponseModel:
```
Ortega-Dan, line 4111: it can be for later on the cleanning but just wanted to make the comment that looks like these 2 models must be unified:
I mean this one, with this one: https://github.com/Nuvalence/dsgov-work-manager/blob/944c7fe12ad0295ba1358e9659fac18d52b061cf/swagger.yaml#L4089-L4089

service/src/main/java/io/nuvalence/workmanager/service/controllers/ConversationsApiDelegateImpl.java
```diff
@@ -132,140 +123,52 @@ public ResponseEntity<ConversationResponseModel> postConversation(
...
     private PageConversationResponseModel createPageConversationResponse(
-            Conversation conversation, List<UUID> participantsIds) {
+            Conversation conversation) {
         PageConversationResponseModel pageConversationResponseModel =
-                conversationMapper.createPageConversationsResponseModel(
-                        conversation, participantsIds);
+                conversationMapper.createPageConversationsResponseModel(conversation);
 
         Message originalMessage =
                 conversation.getReplies().stream()
```
Ortega-Dan, line 174: mm, now I realize this change is even more necessary because the logic in this lines is also done in the ConversationMapper automatically when converting to the ConversationResponseModel, instead of the PageConversactionResponseModel

Ortega-Dan (approved):

service/src/main/java/io/nuvalence/workmanager/service/controllers/ConversationsApiDelegateImpl.java
jcsuarezm: Thank you Dan, I will create a follow-up PR for this.

swagger.yaml
jcsuarezm: Thank you Dan, I will create a follow-up PR for this.

PR #420, opened 2024-02-05
Feature/dsg 2754/get conversation

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/mapper/MessageMapper.java
```diff
@@ -29,6 +29,7 @@ default void createModelToMessageWithSender(
         message.setReadBy(List.of(sender.getId()));
     }
 
+    @Mapping(target = "isOriginalMessage", source = "originalMessage")
```
Ortega-Dan, line 32: This may be just a conversation. But ideally we should use "originalMessage" since json is just the data interchange format, so it should not apply any convention, so the Front-end may also have a isOriginalMessage getter. 
I know we also have the "isComplete", which I believe we could have had it as "completed" instead (and I know the Front-end is already on it).
So if you think it is find to pass it as originalMessage it is fine.

Ortega-Dan: as I was writing this I realize the conversation originalMessage also gets less clear if this is called the same. 
I believe it is fine then.

Ortega-Dan (approved):

service/src/main/java/io/nuvalence/workmanager/service/mapper/MessageMapper.java
jcsuarezm: Thank you Dan, I think either way could be fine, although I prefer isOriginalMessage in the response for clarity. For the time being, I will merge, and as part of the clean-up PR, we could revisit this.

PR #426, opened 2024-02-12
Fix for record update

jcsuarezm:
## Types of changes

Patch
Minor
Major

## Checklist:


helm/work-manager/cerbos-policies/record_resource.yaml
```diff
@@ -3,20 +3,24 @@ resourcePolicy:
...
+
+    - actions: [ "update" ]
+      roles:
+        - record-admin
       effect: EFFECT_ALLOW
 
     - actions: [ "update-admin-data" ]
       roles:
-        - transaction-admin
+        - record-admin
```
Ortega-Dan, line 25: Juan, are these new roles "record-submitter" and "record-admin" going to be included in the tokens ? for now we need to include them to the identity functions.
Even when that is a question, I believe the splitting of this specific roles is not really needed IF things are as it seems in relation for records being dependent on transactions:
based on this validation when creating records: https://github.com/Nuvalence/dsgov-work-manager/blob/8b7a8e1f81f507e0ca590c6800976ed9037b2ba5/service/src/main/java/io/nuvalence/workmanager/service/controllers/RecordsApiDelegateImpl.java#L99-L105
and based on the automatic creation of records when TransactionDefinition and RecordDefinition are linked.
If so, and only if so, I believe keeping just a single set of roles ("transaction-" related ones) is enough. 

This is more to trigger the conversation or evaluation needed before merging this. And if confirmed to be merged, adding the roles to identity functions.

service/src/main/java/io/nuvalence/workmanager/service/controllers/RecordsApiDelegateImpl.java
```diff
@@ -181,26 +179,18 @@ public ResponseEntity<RecordResponseModel> updateRecord(
...
                 mapper.recordToRecordResponseModel(
-                        recordService.updateRecord(recordUpdateRequest, existingRecord)));
+                        recordService.updateRecord(
+                                recordUpdateRequest, id, isAdminForData(userType))));
+    }
+
+    private boolean isAdminForData(String userType) {
+        return userType != null
+                && (userType.equals(UserType.AGENCY.getValue())
+                        || authorizationHandler.isAllowed("update-admin-data", Record.class));
```
Ortega-Dan, line 194: I also believe that these checkings should be done based on Cerbos and ideally not hard code a specific user type. So that those can be configured over cerbos and not require code changes.

helm/work-manager/cerbos-policies/record_resource.yaml
Ortega-Dan: Conversation resolved on the creation of the roles and keeping the roles in order to allow needed granularity.

Ortega-Dan (approved):
I know the agency checking is how it was done before, and there is going to be a ticket to organize all that. 
But I would also recommend the proper location of access based rejection.

service/src/main/java/io/nuvalence/workmanager/service/controllers/RecordsApiDelegateImpl.java
jcsuarezm: Hi Dan, due to issues with the transactional context, this was moved temporarily to the service, but I will be moved back to the controller in a follow-up ticket.

service/src/main/java/io/nuvalence/workmanager/service/service/RecordService.java
```diff
@@ -59,10 +61,27 @@ public Page<Record> getRecordsByFilters(RecordFilters filter) {
...
+                        .orElseThrow(() -> new NotFoundException("Record not found"));
+
+        if (!isAdmin) {
+            if (hasAdminDataChanges(
+                    existingRecord,
+                    updateRequest.getStatus(),
+                    updateRequest.getExpires() != null
+                            ? updateRequest.getExpires().toString()
+                            : null)) {
+                throw new ForbiddenException("User is not authorized to update admin data");
```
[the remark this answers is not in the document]
jcsuarezm: Hi Dan, due to issues with the transactional context, this was moved temporarily to the service, but I will be moved back to the controller in a follow-up ticket.

PR #431, opened 2024-02-14
Dsg-2643: fix code smells

jgnuvalence:
## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/config/WebSecurityConfig.java
```diff
@@ -171,11 +170,6 @@ public WebMvcConfigurer corsConfigurer() {
     @JacocoIgnoreInGeneratedReport(reason = "Simple config class.")
     private class CorsWebMvcConfigurer implements WebMvcConfigurer {
 
-        @Override
```
jgnuvalence, line 174: `setUseTrailingSlashMatchesta` deprecated, removing this seems safe according to my smoke tests, it doesn't look like we're using extra trailing slashes

Ortega-Dan: Not sure about the entire flow to guarantee this is not going to be needed. But is there an alternative non-deprecated approach?

service/src/main/java/io/nuvalence/workmanager/service/camunda/delegates/TransactionDataUpdaterDelegate.java
```diff
@@ -71,14 +71,15 @@ private void setAttributeValue(Transaction transaction, String key, Object value
...
     private Object validateValueType(Transaction transaction, Object value, String attributePath)
             throws IllegalAccessException, InvocationTargetException, NoSuchMethodException {
         PropertyDescriptor descriptor =
                 PropertyUtils.getPropertyDescriptor(transaction, attributePath);
         if (descriptor != null) {
             Class<?> propertyType = descriptor.getPropertyType();
-            if (propertyType.isEnum() && value instanceof String) {
-                value = Enum.valueOf((Class<Enum>) propertyType, (String) value);
+            if (propertyType.isEnum() && value instanceof String valueString) {
+                value = Enum.valueOf((Class<Enum>) propertyType, valueString);
```
Ortega-Dan, line 82: looks like this is a major smell according to SQ. 
If the goal of this PR is just addressing the minor ones then this is likely covered in another ticket. 
Would be awesome if fixable. But feel free to make the call if that goes in another ticket or in this one.

service/src/main/java/io/nuvalence/workmanager/service/config/WebSecurityConfig.java
jgnuvalence: According to the Spring Boot documentation, it does not offer an alternative since it is not safe to take care of this because it generates possible inconsistencies. It is recommended to have consistency in the URLs or handle this with external proxies or internal filters. From my point of view we don't need any of this, we just need to use the correct URLs without extra trailing slashes

service/src/main/java/io/nuvalence/workmanager/service/camunda/delegates/TransactionDataUpdaterDelegate.java
jgnuvalence: I'm taking care of all the code smells since there were not a lot

Ortega-Dan: Ohh, Sorry I was not clear, but according to last SQ report, this is still an open smell:
https://sonarqube.dsgov.demo.nuvalence.io/project/issues?resolved=false&inNewCodePeriod=true&types=CODE_SMELL&branch=dsg-2643%2Ffix-code-smells&id=dsgov-work-manager&open=AY2puANwlHA8G_CZ3UNs

jgnuvalence: oh got it, this is something we need to keep to make the delegate generic, I added a 
@SuppressWarnings("unchecked")

however seems to not work, so I'll mark it as won't resolve, and remove the @SuppressWarnings("unchecked")

PR #441, opened 2024-02-23
dsg-2705 Get first user-tasks and related forms

Ortega-Dan:
https://dol-ewf.atlassian.net/browse/DSG-2705

## Description

Find first user tasks in workflow and provide the applicable forms

## How Has This Been Tested?

API and Unit tests.

Testing configuration is under demo configs branch [dsg-2705/testGettingFirstUserTasks](https://github.com/Nuvalence/dsgov-demo-configurations/tree/dsg-2705/testGettingFirstUserTasks)
image
and subprocess for public user lane:
image
The highlighted tasks are the difference in regards to standard FinancialBenefit demo config.

Under the Following Form Configurations:
(default form is empty)
image

The following results are verified:
Public has only access to user tasks in configured lane:
image
Form configurations and context are considered:
image
Agency has access to both lanes based on panel configuration:
image

## Types of changes

Patch
Minor
Major

## Checklist:


service/src/main/java/io/nuvalence/workmanager/service/utils/camunda/CamundaWorkflowInspector.java
```diff
@@ -75,6 +82,78 @@ public WorkflowTask getWorkflowTask(String taskKey) {
...
+     * @return list of first tasks
+     */
+    public List<WorkflowTask> getWorkflowFirstTasks() {
+
+        Collection<StartEvent> startEvents = modelInstance.getModelElementsByType(StartEvent.class);
+
+        List<UserTask> firstTasks = new ArrayList<>();
+
+        for (StartEvent startEvent : startEvents) {
+            findFirstUserTasks(startEvent, firstTasks, new AtomicBoolean(false));
```
jgnuvalence, line 97: My only question is why use AtomicBoolean since this is not multi-threading as I understand?

jgnuvalence (approved):

service/src/main/java/io/nuvalence/workmanager/service/utils/camunda/CamundaWorkflowInspector.java
Ortega-Dan: Thank you Julian, 
That is right regarding concurrency/multi-threading. The intention of AtomicBoolean here is more on the functionality and clarity of it, because what was needed was a wrapper for a boolean object that could be read by the caller when returned, but avoiding to use the boolean as the return type, and so only set it if needed, instead of sending back not-needed values on every return statement.
Also the need was to preserve the original reference. And a regular Boolean object would not do that because when set it only changes in the scope of the method in which it is, but when the call returns the value in the original Boolean is the one set at the beggining, because it is immutable. 
But AtomicBoolean was a proper and clean mutable wrapper of a boolean value needed in this case.

